PR #10, opened 2019-05-08
Establish Project File Structure

marstr:
This PR seeks to add the most basic skeleton of a Python project. This should ensure that before I begin submitting the [other changes from my fork](https://github.com/marstr/microsoft-authentication-extensions-for-python/tree/providerChain), the basic structure of the project is okay.


setup.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
+
+from setuptools import setup, find_packages
+
+__version__ = "0.0.1"
```
rayluo: If we define version number inside setup.py, we won't be able to [check this package's version on-the-fly](https://www.python.org/dev/peps/pep-0396/#examples). Would you consider [do this](https://stackoverflow.com/a/39671214/728675) insead? <sub>Shameless advertising: I wrote that SO answer.</sub>

marstr: Fixed in 8305767

marstr: Thanks for showing me this! It definitely demonstrates my general noobiness in Python.

setup.py
```diff
@@ -0,0 +1,25 @@
...
+
+setup(
+    name='msal-extensions',
+    version=__version__,
+    package_dir={
+        '': 'src',
+    },
+    packages=find_packages(
+        where="./src",
```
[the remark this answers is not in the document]
marstr: You make a strong argument :) This is fixed in 522be13

setup.py
```diff
@@ -0,0 +1,24 @@
...
+setup(
+    name='msal-extensions',
+    version=__version__,
+    packages=find_packages(),
+    classifiers=[
+        'Development Status :: 2 - Pre-Alpha',
+    ],
+    extra_require={
+        'dev': [
+            'pytest',
```
marstr: Mind confirming whether or not `pytest` is your preferred way to author/execute tests?

setup.py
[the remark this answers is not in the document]
marstr: Note: I realized, one of the reasons I had segmented this into `src` and `tests` was to intrinsically exclude tests from the packages that were discovered by setuptools. To combat this, I added 1a39ab7.

setup.py
rayluo: It is not about noobiness at all! Organizing a python package is a very specific area that most Python developers - who are library consumers - would never touch, because they don't need to.

setup.py
rayluo: I believe there is less a need to standardize the test framework, so I wasn't even paying much attention to this line. :-P You can choose [whatever](https://docs.python-guide.org/writing/tests/) you want.

But then you kind of remind me, this `pytest` line would better NOT be put inside `extraS_require` (you had a typo there, by the way). [The `extras_require` are mean to organize extra FUNCTIONALITY, explained here](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies).

For test dependency, they can be put inside a different parameter, [`tests_require`](https://setuptools.readthedocs.io/en/latest/setuptools.html), so it would look like this: `tests_require=['pytest']`. For example, this is [what PyJWT did](https://github.com/jpadilla/pyjwt/blob/master/setup.py#L72).

rayluo:
Also, I've just created a new `dev` branch right on top of current `master`. Please change your current PR's base branch to `dev`. Thanks!

setup.py
marstr: Another derpy moment! Fixed in 081933c

PR #11, opened 2019-05-14
Prevent entry into a code-block until a file lock can be taken

### LINKED ISSUE #2: Create CrossPlatLock Analog
The .NET library relies on taking write-only locks on a file separate used as a semaphore to ensure that Cache operations do not race each other. That logic is found here: [CrossPlatLock class](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/master/src/Shared/CrossPlatLock.cs)

This is needed before any of the platform specific Token Caches, such as the one represented in #1, can be said to be truly done.

marstr:
This replicates logic in the .NET extensions repository found here:
[{Microsoft Authentication Extensions for .NET}/src/Shared/CrossPlatLock.cs](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/113ecc725891395dd7212aac82d0681f9a0d60ce/src/Shared/CrossPlatLock.cs)

It will allow .NET processes and Python processes to not step on eachother's toes.

It does add a dependency on two new libraries: [psutil](https://pypi.org/project/psutil) and [portalocker](https://pypi.org/project/portalocker)

Closes #2


msal_extensions/_cache_lock.py
```diff
@@ -0,0 +1,41 @@
...
+            os.makedirs(lock_dir)
+
+        self._fh = open(self._lockpath, 'wb+', buffering=0)
+        portalocker.lock(self._fh, portalocker.LOCK_EX)
+        self._fh.write('{} {}'.format(pid, proc.name()).encode('utf-8'))
+
+    def __exit__(self, *args):
+        self._fh.close()
+        try:
+            # Attempt to delete the lockfile. In either of the failure cases enumerated below, it is likely that
```
rayluo: It looks good, but from a feature requirement's standpoint, is such removal really necessary? It seems we can leave it there. The .Net extension seems to also [leave it there](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blame/master/src/Shared/CrossPlatLock.cs#L76-L80).

marstr: Aha, this one I do feel good about! .NET is just sneaky and signals that the file should be deleted when the file handle is first constructed here:
https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/113ecc725891395dd7212aac82d0681f9a0d60ce/src/Shared/CrossPlatLock.cs#L53

For that reason, I believe this should stay as is.

msal_extensions/__init__.py
```diff
@@ -1 +1,3 @@
 __version__ = "0.0.1"
+
+from ._cache_lock import CrossPlatLock
```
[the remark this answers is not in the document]
marstr: Totally fair! Do you have a preference on whether I fix this by hiding it further or making it more public?

Just out of curiosity, could you explain why that's more Pythonic? In general, I've struggled to get my head around the Python idioms around organizing code. My line of thinking for putting it this way is that it forces folks to import it in a single consistent way: `from msal_extensions import CrossPlatLock`. That way if we want to rename the file in the future it's not a breaking change.

msal_extensions/_cache_lock.py
```diff
@@ -0,0 +1,41 @@
...
+    def __enter__(self):
+        pid = os.getpid()
+        proc = psutil.Process(pid)
+        lock_dir = os.path.dirname(self._lockpath)
+        if not os.path.exists(lock_dir):
+            os.makedirs(lock_dir)
+
+        self._fh = open(self._lockpath, 'wb+', buffering=0)
+        portalocker.lock(self._fh, portalocker.LOCK_EX)
+        self._fh.write('{} {}'.format(pid, proc.name()).encode('utf-8'))
```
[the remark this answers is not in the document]
marstr: Love it :) The smaller the number of dependencies, the better in my book. I'll make that change in the morning.

msal_extensions/__init__.py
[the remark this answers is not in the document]
marstr: I guess, by my own reasoning, my preferred option would be to hide it further to prevent avenues for breaking changes in the future. If folks demand access to this helper, we can always make it public later.

msal_extensions/_cache_lock.py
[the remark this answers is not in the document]
marstr: Fixed in 437e032

msal_extensions/__init__.py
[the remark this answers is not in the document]
marstr: Fixed in bda687b

msal_extensions/cache_lock.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class CrossPlatLock(object):
+    """
+    Offers a mechanism for waiting until another process is finished interacting with a shared resource. This is
+    specifically written to interact with a class of the same name in the .NET extensions library.
+    """
+    TIMEOUT = datetime.timedelta(minutes=1)
+    RETRY_WAIT = datetime.timedelta(milliseconds=100)
+    RETRY_COUNT = int(TIMEOUT.total_seconds() / RETRY_WAIT.total_seconds())
```
rayluo: So these part of consts are not used at all. Were you planning to use them in `portalocker.lock(...)`?

msal_extensions/cache_lock.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __init__(self, lockfile_path):
+        self._lockpath = lockfile_path
+
+    def __enter__(self):
+        pid = os.getpid()
+        lock_dir = os.path.dirname(self._lockpath)
+        if not os.path.exists(lock_dir):
+            os.makedirs(lock_dir)
```
rayluo: I can understand your motivation here. But then it is argurably also needed for the real token cache file. So another approach is we move this part of logic to their common caller, the actual TokenCache class in your subsequent PRs.

marstr: Fixed in 1731a6a

msal_extensions/cache_lock.py
marstr: Cruft left-over from when I was trying to implement `portalocker`'s behavior myself! It's easy enough to put back.

This is fixed in aacee74

msal_extensions/_cache_lock.py
rayluo: We had a in-depth offline conversation.

* We believe the motivation of having file removal behavior was partially driven by the convention of writing some debug message into the lock file, which is in-itself technically unnecessary for pure locking purpose.
* And we are aware that the extra file removal behavior here creates chances for racing conditions between file unlock and file removal. And then it forces us to catch and skip those exceptions in this implementation. It won't be necessary if we just use raw locking behavior.

We may revisit this in future.

marstr: We chatted about this a bit offline, and came to the conclusion that there are some benefits to not deleting this file. However, we left deciding that it's an issue we should take up with the .NET behavior and have a continued conversation about it, instead of blocking the PR on it.

PR #14, opened 2019-05-16
Add DP API Cache

### LINKED ISSUE #1: Create a DP API Token Cache (Windows Only)
Create a TokenCache class which extends the existing [SerializableTokenCache](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/f76f3c37c6bf7747971947d300f0a50d3ec701f7/msal/token_cache.py#L204) to handle interacting with a token cache which is persisted in a file protected by Windows provided encryption logic.

The goal is to bring this Python library to parity with the .NET functionality found in the [MsalCacheStorage class.](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/cf9ca48e0295efca0562a743e7758d1c59442f22/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheStorage.cs#L16)

marstr:
Closes #1

### DIFF
--- a/msal_extensions/_windows.py
+++ b/msal_extensions/_windows.py
@@ -1,0 +1,173 @@
+import os
+import ctypes
+from ctypes import wintypes
+import time
+import msal
+import errno
+from .cache_lock import CrossPlatLock
+
+_local_free = ctypes.windll.kernel32.LocalFree
+_memcpy = ctypes.cdll.msvcrt.memcpy
+_crypt_protect_data = ctypes.windll.crypt32.CryptProtectData
+_crypt_unprotect_data = ctypes.windll.crypt32.CryptUnprotectData
+_CRYPTPROTECT_UI_FORBIDDEN = 0x01
+
+
+class DATA_BLOB(ctypes.Structure):
+    """
+    A wrapper for interacting with the _CRYPTOAPI_BLOB type and its many aliases. This type is exposed from Wincrypt.h
+    in XP and above.
+
+    See documentation for this type at:
+    https://msdn.microsoft.com/en-us/7a06eae5-96d8-4ece-98cb-cf0710d2ddbd
+    """
+    _fields_ = [("cbData", wintypes.DWORD), ("pbData", ctypes.POINTER(ctypes.c_char))]
+
+    def raw(self):
+        # type: () -> bytes
+        cb_data = int(self.cbData)
+        pb_data = self.pbData
+        buffer = ctypes.c_buffer(cb_data)
+        _memcpy(buffer, pb_data, cb_data)
+        _local_free(pb_data)
+        return buffer.raw
+
+
+# This code is modeled from a StackOverflow question, which can be found here:
+# https://stackoverflow.com/questions/463832/using-dpapi-with-python
+class WindowsDataProtectionAgent(object):
+    def __init__(self, **kwargs):
+        # type: (str) -> None
+        if 'entropy' in kwargs:
+            entropy = kwargs['entropy'].encode('utf-8')
+            self._entropy_buffer = ctypes.c_buffer(entropy, len(entropy))
+            self._entropy_blob = DATA_BLOB(len(entropy), self._entropy_buffer)
+        else:
+            self._entropy_buffer = None
+            self._entropy_blob = None
+
+    def protect(self, message):
+        # type: (str) -> bytes
+
+        message = message.encode('utf-8')
+        message_buffer = ctypes.c_buffer(message, len(message))
+        message_blob = DATA_BLOB(len(message), message_buffer)
+        result = DATA_BLOB()
+
+        if self._entropy_blob:
+            entropy = ctypes.byref(self._entropy_blob)
+        else:
+            entropy = None
+
+        if _crypt_protect_data(
+                ctypes.byref(message_blob),
+                u"python_data",
+                entropy,
+                None,
+                None,
+                _CRYPTPROTECT_UI_FORBIDDEN,
+                ctypes.byref(result)):
+            return result.raw()
+        return b''
+
+    def unprotect(self, cipher_text):
+        # type: (bytes) -> str
+        ct_buffer = ctypes.c_buffer(cipher_text, len(cipher_text))
+        ct_blob = DATA_BLOB(len(cipher_text), ct_buffer)
+        result = DATA_BLOB()
+
+        if self._entropy_blob:
+            entropy = ctypes.byref(self._entropy_blob)
+        else:
+            entropy = None
+
+        if _crypt_unprotect_data(
+            ctypes.byref(ct_blob),
+            None,
+            entropy,
+            None,
+            None,
+            _CRYPTPROTECT_UI_FORBIDDEN,
+            ctypes.byref(result)
+        ):
+            return result.raw().decode('utf-8')
+        return u''
+
+
+class _WindowsTokenCache(msal.SerializableTokenCache):
+    DEFAULT_CACHE_LOCATION = os.path.join(os.getenv('LOCALAPPDATA'), '.IdentityService', 'msal.cache')
+    DEFAULT_ENTROPY = ''
+
+    def __init__(self, **kwargs):
+        super(_WindowsTokenCache, self).__init__()
+
+        self._cache_location = _WindowsTokenCache.DEFAULT_CACHE_LOCATION  # type: str
+        if 'cache_location' in kwargs:
+            self._cache_location = kwargs['cache_location'] or _WindowsTokenCache.DEFAULT_CACHE_LOCATION
+        self._lock_location = self._cache_location + '.lockfile'
+
+        entropy = _WindowsTokenCache.DEFAULT_ENTROPY
+        if 'entropy' in kwargs:
+            entropy = kwargs['entropy']
+        self._dp_agent = WindowsDataProtectionAgent(entropy=entropy)
+        self._last_sync = 0  # _last_sync is a Unixtime
+
+    def _has_state_changed(self):
+        # type: () -> Bool
+        try:
+            return self.has_state_changed or self._last_sync < os.path.getmtime(self._cache_location)
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise e
+            return False
+
+    def add(self, event, **kwargs):
+        if self._has_state_changed:
+            try:
+                self._read()
+            except OSError as e:
+                if e.errno != errno.ENOENT:
+                    raise e
+        super(_WindowsTokenCache, self).add(event, **kwargs)
+        self._write()
+
+    def update_rt(self, rt_item, new_rt):
+        if self.has_state_changed:
+            try:
+                self._read()
+            except OSError as e:
+                if e.errno != errno.ENOENT:
+                    raise e
+        super(_WindowsTokenCache, self).update_rt(rt_item, new_rt)
+        self._write()
+
+    def remove_rt(self, rt_item):
+        if self._has_state_changed:
+            try:
+                self._read()
+            except OSError as e:
+                if e.errno != errno.ENOENT:
+                    raise e
+        super(_WindowsTokenCache, self).remove_rt(rt_item)
+        self._write()
+
+    def find(self, credential_type, target=None, query=None):
+        if self._has_state_changed:
+            try:
+                self._read()
+            except OSError as e:
+                if e.errno != errno.ENOENT:
+                    raise e
+        return super(_WindowsTokenCache, self).find(credential_type, target=target, query=query)
+
+    def _write(self):
+        with CrossPlatLock(self._lock_location), open(self._cache_location, 'wb') as fh:
+            fh.write(self._dp_agent.protect(self.serialize()))
+        self._last_sync = int(time.time())
+
+    def _read(self):
+        with CrossPlatLock(self._lock_location), open(self._cache_location, 'rb') as fh:
+            cipher_text = fh.read()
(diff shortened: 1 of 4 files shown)


msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+            self._cache_location = kwargs['cache_location'] or _WindowsTokenCache.DEFAULT_CACHE_LOCATION
+        self._lock_location = self._cache_location + '.lockfile'
+
+        entropy = _WindowsTokenCache.DEFAULT_ENTROPY
+        if 'entropy' in kwargs:
+            entropy = kwargs['entropy']
+        self._dp_agent = WindowsDataProtectionAgent(entropy=entropy)
+        self._last_sync = 0  # _last_sync is a Unixtime
+
+    def _has_state_changed(self):
```
rayluo: My bad. I used the name `has_state_changed` in `SerializableTokenCache`. At a hindsight, I would probably rename it to `has_memory_state_changed` soon.
And then your function here should really be `has_file_state_changed()`. And these 2 flags are exactly the opposite meaning.

msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+        # type: () -> Bool
+        try:
+            return self.has_state_changed or self._last_sync < os.path.getmtime(self._cache_location)
```
rayluo: Now you might want to skip the `self.has_memory_state_changed` check here, because ideally it shouldn't happen in here in this class.

msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise e
+            return False
+
+    def add(self, event, **kwargs):
+        if self._has_state_changed:
```
rayluo: You mean `self._has_state_changed`**()**?

msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+            try:
+                self._read()
+            except OSError as e:
+                if e.errno != errno.ENOENT:
+                    raise e
+        super(_WindowsTokenCache, self).add(event, **kwargs)
+        self._write()
+
+    def update_rt(self, rt_item, new_rt):
```
rayluo: I know you were not aware of this: we've been working on a new PR in MSAL Python which would introduce a bunch of new `remove_xx(...)` methods, but here comes the good news! We [optimized it](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/41/files#diff-6f83eec40714321407e16ebed558f014R287) for you, so that you would only need to wrap one `modify()` instead of all those `update_xx()` and `remove_xx()`. You would need to change these 2 methods here, after we release MSAL Python 0.4.x, soon.

msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+        if self.has_state_changed:
+            try:
+                self._read()
+            except OSError as e:
+                if e.errno != errno.ENOENT:
+                    raise e
+        super(_WindowsTokenCache, self).update_rt(rt_item, new_rt)
+        self._write()
+
+    def remove_rt(self, rt_item):
```
rayluo: Same as above `update_rt()`.

msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+        if self._has_state_changed:
+            try:
+                self._read()
+            except OSError as e:
+                if e.errno != errno.ENOENT:
+                    raise e
+        super(_WindowsTokenCache, self).remove_rt(rt_item)
+        self._write()
+
+    def find(self, credential_type, target=None, query=None):
```
rayluo: This time though, we would want to use **kwargs to accomodate possible future extra values introduced in the base class.

```python
def find(self, credential_type, **kwargs):
    ...
    return super(...).find(credential_type, **kwargs)
```

msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+        if self._has_state_changed:
```
rayluo: You mean `self._has_state_changed`**()**?

msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    _fields_ = [("cbData", wintypes.DWORD), ("pbData", ctypes.POINTER(ctypes.c_char))]
+
+    def raw(self):
+        # type: () -> bytes
+        cb_data = int(self.cbData)
+        pb_data = self.pbData
+        buffer = ctypes.c_buffer(cb_data)
+        _memcpy(buffer, pb_data, cb_data)
+        _local_free(pb_data)
```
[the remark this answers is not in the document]
marstr: So, the local_free is definitely necessary as we don't want to create a memory leak! However, you're totally right that it should be safe to call this method twice in a way it just isn't right now. I'll fix that up.

msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+            None,
+            None,
+            _CRYPTPROTECT_UI_FORBIDDEN,
+            ctypes.byref(result)
+        ):
+            return result.raw().decode('utf-8')
+        return u''
+
+
+class _WindowsTokenCache(msal.SerializableTokenCache):
```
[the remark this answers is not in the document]
marstr: Fixed in 91c6badd2

msal_extensions/_windows.py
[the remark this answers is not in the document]
marstr: Fixed in b9becbcd8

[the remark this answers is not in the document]
marstr: So, I'm going to have to walk this change back. I got to a Windows machine and ran the tests, and curiously pulling this object out into the destructor causes a seg-fault that crashes the host Python process with it.

msal_extensions/_windows.py
marstr: Fixed in 9ef612f

msal_extensions/_windows.py
marstr: Fixed in 9ef612f

msal_extensions/_windows.py
marstr: Hey, this is great news! If you don't mind, I'd like to keep this PR scoped to just getting this Windows agent targeting msal~=0.3.0. I can submit another PR to adopt 0.4.0.

msal_extensions/_windows.py
marstr: Same as here: https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/14#discussion_r286144472

msal_extensions/_windows.py
marstr: Fixed in 9ef612f

msal_extensions/_windows.py
marstr: I went ahead and renamed this method to `needs_refresh`, what do you think of that name?

msal_extensions/_windows.py
marstr: I believe this is feedback I should react to at the same time that I respond to this thread: https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/14#discussion_r286144472

msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+        buffer = ctypes.c_buffer(cb_data)
+        _memcpy(buffer, pb_data, cb_data)
+        _local_free(pb_data)
+        return buffer.raw
+
+
+# This code is modeled from a StackOverflow question, which can be found here:
+# https://stackoverflow.com/questions/463832/using-dpapi-with-python
+class WindowsDataProtectionAgent(object):
+    def __init__(self, **kwargs):
```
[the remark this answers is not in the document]
marstr: Fixed in 082684b

Thanks for this by the way, I like the refactored version a lot better.

msal_extensions/_windows.py
```diff
@@ -0,0 +1,173 @@
...
+class _WindowsTokenCache(msal.SerializableTokenCache):
+    DEFAULT_CACHE_LOCATION = os.path.join(os.getenv('LOCALAPPDATA'), '.IdentityService', 'msal.cache')
+    DEFAULT_ENTROPY = ''
+
+    def __init__(self, **kwargs):
+        super(_WindowsTokenCache, self).__init__()
+
+        self._cache_location = _WindowsTokenCache.DEFAULT_CACHE_LOCATION  # type: str
+        if 'cache_location' in kwargs:
+            self._cache_location = kwargs['cache_location'] or _WindowsTokenCache.DEFAULT_CACHE_LOCATION
```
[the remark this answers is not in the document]
marstr: Fixed in a712a73

msal_extensions/_windows.py
rayluo: Sure. This conversation was mean to be a FYI and not actionable until MSAL Python 0.4.0 is released.

Yet another interesting FYI: You might already know, that `msal~=0.3.0` notation means [automatically adopt every future versions with same major version number](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) rather than "0.3.0 only". It is the desirable notation and we use it all the time. Just keep in mind that this time we would want to manually patch it asap, AFTER msal python 0.4.0. :-)

I'll close this comment soon.

msal_extensions/_windows.py
rayluo: This one is about such `find(...)` itself, so it has nothing to do with the upcoming new `modify(...)`. It is just a future-proof notation to make this `find(...)` forward-compatible with any new keyworded parameter(s) introduced into base class's `find(...)` in the future. I would still suggest we do it right this time, once and for all. :-)

msal_extensions/_windows.py
marstr: > Yet another interesting FYI: You might already know, that msal~=0.3.0 notation means automatically adopt every future versions with same major version number rather than "0.3.0 only". It is the desirable notation and we use it all the time. Just keep in mind that this time we would want to manually patch it asap, AFTER msal python 0.4.0. :-)

I just tested it out by creating a new venv and running `pip install tropo~=0.1.0`, I confirmed that the behavior seems to be special-cased for the '0.x' major version band to treat minor versions as breaking. i.e installing tropo~=0.1.0 resolves to 0.1.4, not 0.2.1 even though [it's available](https://pypi.org/project/tropo/#history).

msal_extensions/_windows.py
rayluo: You ask for it! :-)

Per your request, I think the new name `needs_refresh` has the same mistake I made when I came up with the `has_state_changed` in the base class. They are both arguably vague in terms of direction. Should it be a need to handle 2-way sync in the future, the current name would become insufficient. That being said, the point is not to predict the future, because you can't. So, don't bother. :-)  Just one minor thing though. If my aforementioned mistake teaches me anything, it is that this `needs_refresh` behavior should largely be internal. So if you add a `_` prefix to make it `_needs_refresh`, it would be perfect because you can rename it later easily.

PS: Naming inside this WindowsTokenCache class matters, because it would become one of the major API of this extension module, even though we might have a factory for it.

marstr: Fixed in e4132a1

msal_extensions/_windows.py
marstr: Fixed in 61104a9

msal_extensions/_windows.py
marstr: Oh, maybe we're talking about different things! Sorry, I read that as "As it is, you'll accidentally be pulling in 0.4.0", what you probably meant was "In order to fix this in the next PR, you'll need to remember to bump this version number."

rayluo: Thanks for your experiment! I actually did not know that `~=` notation is "special-cased for the '0.x' major version band to treat minor versions as breaking. i.e installing tropo~=0.1.0 resolves to 0.1.4, not 0.2.1 even though it's available". That is a surprising "feature" to me.

Previously I tend to use the more explicit notation such as ">=0.1, <1" whose behaviour matches my expectation and mathematics: gives a highest available version within that range.

msal_extensions/_windows.py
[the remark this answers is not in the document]
rayluo: If that is the case, can we have one more try as to move that `_local_free()` out of `raw()`, and put it into `protect()` in this manner? On paper, it should work.

```python
    def protect(self, ...):
        ...
        try:
            return result.raw()
        finally:
            _local_free(result.pbData)
```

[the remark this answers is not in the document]
marstr: Love it :) I'll give it another whack in the morning.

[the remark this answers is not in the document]
marstr: So the trick ended up being that LocalFree **MUST NOT** be called memory allocated by ctypes.create_string_buffer, but MUST be called on memory allocated by the CryptProtectData and CryptUnprotectData functions. It just so happened that I only ever called `raw` on the latter, and never called `raw` for the former. This means that using `try` & `finally` allows us to manage the lifetime of only specific structs, where as moving it into a `__del__` method was breaking the contract.

Long story short, this is fixed in  161e6cf. On the otherhand, a4efa654 can be ignored.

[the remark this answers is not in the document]
rayluo: Thanks for doing all the hard work to validate our latest fix, and documeting your finding! For the record, even if that 80e735c would work, I would still prefer the 2a25b5f approach. Because that's fundamentally the right pattern to manage memory (came from my 7 years on C/C++).

🚢

PR #15, opened 2019-05-16
Add KeyChain Cache, Plain-Text File Cache, and Cache Factory

### LINKED ISSUE #3: Create a KeyChain Token Cache (MacOS Only)
In order to store Access Tokens in a secure way on MacOS, we should lean on KeyChain as the .NET library does with the [MacKeyChain class](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/cf9ca48e0295efca0562a743e7758d1c59442f22/src/Shared/MacKeyChain.cs#L24).

In practice, this means implementing a class that extends [SerializableTokenCache](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/f76f3c37c6bf7747971947d300f0a50d3ec701f7/msal/token_cache.py#L204) which encapsulates the KeyChain interactions.

Before this issue can truly be considered done, #2 must be finished.

### LINKED ISSUE #5: Create Secret Backend Mux
Folks should have a platform-independent interface for getting the most appropriate Token Cache. This will save customers from needing to understand that they are choosing between implementations of types noted in #1, #3, or #4. This likely just looks like a function named `get_protected_token_cache` or similar.

marstr:
Closes #3
Closes #5

I usually like to keep my PRs small. This time I got tantalized into exploding the size to do some integration test with the Azure CLI.

We are rapidly approaching functional parity with my initial prototype, but with much much cleaner code.


msal_extensions/token_cache.py
```diff
@@ -0,0 +1,17 @@
...
+
+def get_protected_token_cache(enforce_encryption=False, **kwargs):
+    if sys.platform.startswith('win'):
+        return WindowsTokenCache(**kwargs)
+    elif sys.platform.startswith('darwin'):
+        return OSXTokenCache(**kwargs)
+    elif enforce_encryption:
+        raise RuntimeError('no protected token cache for platform {}'.format(sys.platform))
+    else:
+        raise NotImplementedError('No fallback TokenCache is implemented yet.')
```
marstr: A future PR will have this implemented. But for the sake of having each PR add ONE thing, I didn't add that implementation.

tests/test_osx_token_cache.py
```diff
@@ -1,19 +0,0 @@
...
-else:
-    from msal_extensions.osx import KeychainItemNotFoundError
-    from msal_extensions.token_cache import OSXTokenCache
-
-
-def test_read_cache():
-    try:
-        subject = OSXTokenCache()
-        tokens = subject.find(msal.TokenCache.CredentialType.ACCESS_TOKEN)
-        assert len(tokens) > 0
```
marstr: This line was causing CI to fail, because the failure characteristics for MacOS changed here from raising an exception to just returning no token. However, rewriting this test it became clear that it was either "we can definitely say this passed, or we should just skip it." Such a test felt redundant enough to me that I just deleted this.

msal_extensions/token_cache.py
```diff
@@ -0,0 +1,196 @@
...
+    """Creates a directory, and any necessary parents.
+
+    This implementation based on a Stack Overflow question that can be found here:
+    https://stackoverflow.com/questions/600268/mkdir-p-functionality-in-python
+
+    If the path provided is an existing file, this function raises an exception.
+    :param path: The directory name that should be created.
+    """
+    try:
+        os.mkdir(path)
```
rayluo: I believe you mean to use [`os.makedirs()`](https://docs.python.org/2/library/os.html#os.makedirs)?

msal_extensions/token_cache.py
```diff
@@ -0,0 +1,196 @@
...
+        For types derived of this, class that will be a file, which has the ability to track a last
+        modified time.
+        """
+        with open(self._cache_location, 'wb') as handle:
+            handle.write(self.serialize())
+
+    def _read(self):
+        """Fetches the contents of a file and invokes deserialization."""
+        with open(self._cache_location, 'rb') as handle:
+            contents = handle.read()
```
rayluo: I know this part is just a copy&paste from our previously reviewed implementation in windows.py. Now on a second thought, perhaps we can avoid all those duplicated (thrice) `try: ... except IOError: ...` logic from its callers, by adjust this method's implementation (and semantics) into this?

```python
def _read(self):
    try:
        with open(...) as handle:
            return handle.read()
    except IOError as exp:
        if exp.errno != errno.ENOENT:
            raise
```

Note: It is true that this new proposed alternative does NOT do `deserialize(...)` inside, thus push that workload to the callers to do `self.deserialize(self._read())` (which IMHO is semantically clearer because `_read()` has return value thus behaves more like a `read()`), but the real benefit would be the subclass `WindowsTokenCache._read()` can then inherit that exception handling logic here, thus becomes:

```python
class WindowsTokenCache(FileTokenCache):
    ...
    def _read():
        return self._dp_agent.unprotect(super(WindowsTokenCache, self)._read()))
```

msal_extensions/token_cache.py
```diff
@@ -0,0 +1,196 @@
...
+                try:
+                    self._read()
+                except IOError as exp:
+                    if exp.errno != errno.ENOENT:
+                        raise exp
+            return super(FileTokenCache, self).find(credential_type, **kwargs)
+
+    def __getattr__(self, item):
+        # Instead of relying on implementers remembering to update _last_sync, just detect that one
+        # of the relevant methods has been called take care of it for derived types.
```
rayluo: Thank you for paying attention to this kind of details, and coming up with a creative solution!

On the other hand, how about consider NOT do this `__getattr__(...)` magic (which even pylint complains for), and do an old-school inheritance instead?

```python
def serialize():
    self._last_sync = int(time.time())  # In fact, this is the more precise spot to do it, i.e. this way you won't update the time if an IOError would happen
    return super(..., self).serialize()
    # And similar approach to deserialize()
```

msal_extensions/token_cache.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class WindowsTokenCache(FileTokenCache):
+    """A SerializableTokenCache implementation which uses Win32 encryption APIs to protect your
+    tokens.
+    """
+    def __init__(self,
+                 cache_location=None,
+                 lock_location=None,
+                 entropy=''):
```
rayluo: I would suggest to NOT duplicate `cache_location` and `lock_location` here, because they are semantically NOT used by this sub-class anyway. The following approach would also be future-proof if the base class would invent some new parameters in the future.

```python
    def __init__(self, entropy="", **kwargs):
        super(..., self).__init__(**kwargs)
        self._dp_agent = WindowsDataProtectionAgent(entropy=entropy)
```

msal_extensions/token_cache.py
```diff
@@ -0,0 +1,196 @@
...
+class OSXTokenCache(FileTokenCache):
+    """A SerializableTokenCache implementation which uses native Keychain libraries to protect your
+    tokens.
+    """
+
+    def __init__(self,
+                 cache_location='~/.IdentityService/msal.cache',
+                 lock_location=None,
+                 service_name='Microsoft.Developer.IdentityService',
+                 account_name='MSALCache'):
```
rayluo: This one is slightly different than the `WindowsTokenCache.__init__(...)`. The `cache_location` parameter does need to be explicitedly (re)defined here because it would have a different default value. But the `lock_location` can probably still go with the `**kwargs` pattern.

msal_extensions/token_cache.py
marstr: Nice catch! Fixed in d3059c37617d75d1c583615fc65ad95e3036a99e

msal_extensions/token_cache.py
marstr: I've been harboring some reservations about this magic myself! I like the idea of keeping `_last_sync` out of the `_read` and `_write` implementations altogether. However, I'd like to keep `serialize` and `deserialize` as free of side-effects as possible. For that reason I'll move it into `modify`, `find`, and `add` which innately have side effects! The other thing I realized over the last couple of days, is that it's a lot more reliable to use the last modified time from the file rather than capturing that time via the Python system timer. While I have the hold on the CrossPlatLock, I'll read that metadata and use that for the value of `_last_sync`

rayluo: I really appreciate your conscious effort on keeping the `serialize()` and `deserialize()` free of side effect. [Their current status](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/0.4.0/msal/token_cache.py#L292-L304) is indeed [as simple as possible, and that wasn't easy](https://www.azquotes.com/picture-quotes/quote-there-are-two-ways-of-constructing-a-software-design-one-way-is-to-make-it-so-simple-tony-hoare-54-50-35.jpg). That being said, if we consider the "last_sync" time as an intrinsic characteristic of the serialization/deserialization itself, it is OK to record it there, by just one extra line of code.
But again, we are open if you have different ideas. :-)

marstr: Fixed in 1dff899c2c552e7c9b9442432fbeac981b488fac

msal_extensions/token_cache.py
```diff
@@ -160,15 +156,14 @@ def __init__(self,
             lock_location=lock_location)
         self._dp_agent = WindowsDataProtectionAgent(entropy=entropy)
 
-    def _write(self):
+    def _write(self, contents):
         with open(self._cache_location, 'wb') as handle:
```
rayluo: Is there any specific reason that this method does not inherit from its base class? Currently you lost that special exception handling behavior.

msal_extensions/token_cache.py
```diff
@@ -160,15 +156,14 @@ def __init__(self,
...
-            handle.write(self._dp_agent.protect(self.serialize()))
+            handle.write(self._dp_agent.protect(contents))
 
     def _read(self):
         with open(self._cache_location, 'rb') as handle:
```
rayluo: Again, by not inheriting the base method, we lost the customized exception handling logic here.

A different open mode seems not be a reason. In fact, Travis CI is also complaining the "rs" mode in the base class. :-)

msal_extensions/osx.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+    return '/System/Library/Frameworks/{0}.framework/{0}'.format(name)
+
+
+# Load native MacOS libraries
+try:
+    _SECURITY = _ctypes.CDLL(_get_native_location('Security'))
+    _CORE = _ctypes.CDLL(_get_native_location('CoreFoundation'))
+except OSError:
+    _SECURITY = None
```
rayluo: If I understand this correctly, when we "recover" from OSError and assign `_SECURITY` as `None`, the code below `SOMETHING = _SECURITY.SomeThing` would still blow up. So, shall we just do a logging here or - my personal favorite - not catching this OSError exception at all?

msal_extensions/osx.py
```diff
@@ -0,0 +1,315 @@
...
+            'see https://opensource.apple.com/source/CarbonHeaders/CarbonHeaders-18.1/MacErrors.h'\
+                .format(self.exit_status)
+
+
+class KeychainAccessDeniedError(KeychainError):
+    """The exceptions that's raised when a Keychain is set to be queried, but the current process
+    has not been give permission to access it, and a query comes anyway."""
+    EXIT_STATUS = -128
+
+    def __init__(self):
```
rayluo: So the first 70+ lines here essentially implement an exception hierarchy to represent the flat underlying error codes. While such motivation is understandable (so that the downstream developers would only need to catch a descriptive exception, not needing to know and compare a specific errno value), the real benefit seems only meaningful when/if a downstream developer would specifically catch and handle a particular subset of these exceptions but not the others. Otherwise, there is no difference than always raising the same base `KeychainError` which already contains the `exit_status` and, to Martin's credit, a hint pointing to Apple's official single source of truth. Downstream developer's life would be exactly the same when we raise only `KeychainError` and not its children.

If, however, you still insist to do the exception tree here, that's also fine. In such case the suggestion would become more tactical. We should keep the same `__init__(self, exit_status)` here (otherwise we are violating the L of [SOLID principle](https://en.wikipedia.org/wiki/SOLID)). And then in this particular case, that means we can completely remove a duplicated `__init__(...)` implementation here.

msal_extensions/osx.py
```diff
@@ -0,0 +1,315 @@
...
+        super(KeychainAccessDeniedError, self).__init__(KeychainAccessDeniedError.EXIT_STATUS)
+
+
+class NoSuchKeychainError(KeychainError):
+    """The exception that's raised when a Keychain is set to be queried, but it is non-existent, and
+    a query comes anyway."""
+    EXIT_STATUS = -25294
+
+    def __init__(self, name):
```
rayluo: Following the conversation above, the tactical suggestion here would be change this constructor to:

```python
def __init__(exit_status, name):
    super(NoSuchKeychainError, self).__init__(exit_status)
    self.name = name
```

You may wonder, "wouldn't this constructor expose the internal const EXIT_STATUS and then makes it a burden for the caller to provide such value"? The fact is, the caller of this constructor is our own code `_construct_error()`, which would already need to know the mapping between those exit_status and their corresponding classes anyway.

Regarding to downstream app developers, they won't need to call these constructors, and more importantly, we do not (and can not) completely shield them from knowing the exit_status anyway. They still need to know them and then look it up from Apple's source.

msal_extensions/osx.py
```diff
@@ -0,0 +1,315 @@
...
+def _construct_error(exit_status, **kwargs):
+    if exit_status == KeychainAccessDeniedError.EXIT_STATUS:
+        return KeychainAccessDeniedError()
+    if exit_status == NoSuchKeychainError.EXIT_STATUS:
+        return NoSuchKeychainError(**kwargs)
+    if exit_status == NoDefaultKeychainError.EXIT_STATUS:
+        return NoDefaultKeychainError()
+    if exit_status == KeychainItemNotFoundError.EXIT_STATUS:
+        return KeychainItemNotFoundError(**kwargs)
+    return KeychainError(exit_status)
```
rayluo: Syntactic sugar:

```python
def _construct_error(exit_status, **kwargs):
    return {
        KeychainAccessDeniedError.EXIT_STATUS: KeychainAccessDeniedError,
        ...
        }.get(exit_status, KeychainError)(exit_status, **kwargs)
```

msal_extensions/token_cache.py
marstr: I realized that the default value is actually the same as the default value in FileTokenCache. For that reason, the cache_location value specified here is unnecessary.

Fixed in 0327141ce3e32968c30e06ef722e25849782409a

msal_extensions/token_cache.py
marstr: Fixed in 0327141

msal_extensions/token_cache.py
marstr: Fixed the `"rs"` thing in 9c59a72e9e737a711d0a17f64bb39225235d520a and added a test to get coverage of 9f459fb74fc5c2f5dd0952cf6b41e45368802833

msal_extensions/token_cache.py
marstr: So, I've been thinking about it, and I think I'd rather let _read and _write throw exceptions if they're not able to complete their function. The caller (`add`, `modify`, and `find` in our case) can decide whether or not it was acceptable to not not be able to perform that function. This change made here: 47e985c899fbfbab9a3f70db3b81a873666c5307

To your broader point though, I think that `FileTokenCache` is the only implementation that writing in plain text UTF-8. I don't think there's much opportunity for code re-use by encouraging folks to call `super` here. Seems like we may just be on different wave-lengths on this. Let me know if there's something I'm missing!

msal_extensions/token_cache.py
marstr: Contract changed to matched the proposed one here: 1dff899c2c552e7c9b9442432fbeac981b488fac

More conversation about the error handling here: https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/15#discussion_r292203890

msal_extensions/token_cache.py
rayluo: Thanks. Just one more thing. You now have a typo of [duplicated line](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/15/files#diff-8e370721e03e7dd603efc5fe6fa2852aR160). :-)

msal_extensions/token_cache.py
rayluo: Really LOVE this kind of conversation! :-)  I'm going to agree on your latest implementation, meanwhile I'm still providing some thoughts for the sake of mutual learning.

Regarding to putting those exception handling logic out of _read() and _write() and move to their callers, I understand your point and I agree. In 98% of time it would be the correct choice because exception handling should happen more close to the "application" layer, rather than in the "underlying common tool" layer. In this particular case, though, it might fall into the 1% of time when it doesn't really matter, because those two _read() and _write() are not really a common reusable utility to the outside world, they are only needed inside this `FileTokenCache` class family, so it would be fine for them to include some behaviors common to this `FileTokenCache` family, if that means better code reuse. All in all, since the difference here is not that significant and "it doesn't matter", let's settle this case and just use your lastest implementation. :-)

Regarding to that broader point of method inheritance, I know the actual code re-use benefit is not significant in this particular case (but then again, we introduced one more place for that potential "rs" thing :-) ). It is just more of a conceptual convention that a sub-class method is supposed to inherit its base class behavior and build on top of that, whenever possible. Imagine a new maintainer is picking up this method, his first reaction would probably be alerted: "Why does this sub-class method not inherit from the base method? Given that there was no inline comment to confirm this unusual behavior is intended, could this be a bug?", and he would need an investigation to understand what else, if any, is in the base class that could be missing in the subclass; or is it really an intended behavior. All of those were indeed what I did when I was doing this PR review; I would not have to do so if this `WindowsTokenCache._read()` was implemented as a logically self-contained `return self._dp_agent.unprotect(super(..., self)._read())`. (And, for the record, I did same investigation on `OSXTokenCache._read()` and understood why it did not inherit from the base. So that part is fine.)

Resolving this conversation now. :-)

msal_extensions/token_cache.py
marstr: Derp! Fixed in e6009cdd5bdcc9c25282ee79392951b4114f0786

msal_extensions/osx.py
marstr: Oooh, I love this. I would argue that this is not a violation of the 'L' from SOLID. 

Basically, my position is that adding the argument `exit_status` here would be misleading, as the only acceptable value for an AccessDeniedError is `-128`. Further more, I believe Liskov substitution is only talking about the properties of instantiated objects. At least in other languages, I haven't run into examples of needing to have covariant or contravariant constructors! Let me know if Python is doing something unknown to me that changes my thinking. :)

msal_extensions/osx.py
marstr: That looks a lot better! And honestly given the argument above, that this library is the only one that will ever actually construct one of these errors, I think it is enough prettier that it warrants the misleading nature of the additional parameter.

msal_extensions/osx.py
marstr: Wow alright, well I don't think my argument above is wrong! However, seeing the [syntax sugar](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/15/files/d3059c37617d75d1c583615fc65ad95e3036a99e#r292215210) your approach unlocks, I think it warrants having a slightly misleading constructor. :) Especially as you say [here](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/15/files/d3059c37617d75d1c583615fc65ad95e3036a99e#r292214519), we'll be the only ones to ever have to actually invoke it. My strong position was weakly held, and I am yielding it!

marstr: Fixed in 2962bd8d69a24544a0f6f725e7e0e89cb16fbcea

msal_extensions/osx.py
marstr: Fixed in 2962bd8d69a24544a0f6f725e7e0e89cb16fbcea

msal_extensions/osx.py
marstr: Fixed in 2962bd8d69a24544a0f6f725e7e0e89cb16fbcea

msal_extensions/osx.py
marstr: Note, I did add `assert` entries. This way we have some certainty that the correct value is being injected, plus people can [invoke python in such a way that these assertions disappear anyway](https://docs.python.org/3/using/cmdline.html#cmdoption-o).

msal_extensions/osx.py
marstr: I'm guessing I did this before I guarded my import statements by platform. 😝

marstr: Fixed in 469da0073da2b480ec5ce9a2e78165cfa0b96cee

marstr:
Got word from @rayluo offline that there were still some outstanding problems with the Error factory. I've decided to just  go ahead and consolidate the different Error Types into one. It's easier to add to public surface area than remove, after all.

msal_extensions/osx.py
rayluo: Resolving this conversation after [offline discussion](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/15#issuecomment-505079762)

msal_extensions/osx.py
rayluo: Resolving this conversation after [offline discussion](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/15#issuecomment-505079762)

msal_extensions/osx.py
rayluo: Resolving this conversation after [offline discussion](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/15#issuecomment-505079762)

msal_extensions/token_cache.py
```diff
@@ -0,0 +1,193 @@
...
+if sys.platform.startswith('win'):
+    from .windows import WindowsDataProtectionAgent
+elif sys.platform.startswith('darwin'):
+    from .osx import Keychain
+
+
+def get_protected_token_cache(
+        cache_location=None,
+        lock_location=None,
+        enforce_encryption=False, **kwargs):
```
rayluo: From a usability standpoint, wouldn't it be somewhat misleading that such constructor's name is `get_protected_token_cache(...)`, yet one of the optional parameter is `enforce_encryption=False`? It would cause confusion for dev who reads our api signature and even its docs, but not reading its implementation (which means the majority of devs), "am I supposed to always enable `enforce_encryption` or what?".

Would better make the `encryption_thing=True` the default behavior. And then of course we would probably want to NOT blow out by `RuntimeError`. A noticeable warning would be fine. Please see my another comment for a more detailed proposal.

msal_extensions/token_cache.py
```diff
@@ -0,0 +1,193 @@
...
+import msal
+from .cache_lock import CrossPlatLock
+
+if sys.platform.startswith('win'):
+    from .windows import WindowsDataProtectionAgent
+elif sys.platform.startswith('darwin'):
+    from .osx import Keychain
+
+
+def get_protected_token_cache(
```
rayluo: What if we skip this helper completely, and just define the following into the `__init__.py`? It would be conceptually one less layer of helper, from the perspective of app devs. And it is [believed to be](https://stackoverflow.com/a/791107) the [pattern from the python os package](https://github.com/python/cpython/blob/v3.8.0b1/Lib/os.py#L48-L89).

```python
if sys.platform.startswith('win'):
    from .token_cache import WindowsTokenCache as TokenCache
elif sys.platform.startswith('darwin'):
    from .token_cache import OSXTokenCache as TokenCache
else:
    from .token_cache import UnencryptedTokenCache as TokenCache
```

And then the`UnencryptedTokenCache` (formerly the  `FileTokenCache`) constructor can have that "no encryption" warning behavior built-in, even without a dedicated flag for it.

```python
class UnencryptedTokenCache(...):
    def __init__(self, ...):
        ...
        warnings.warn("You are using an unprotected token cache for platform {}".format(sys.platform)
        # Note: since this becomes the default behavior now,
        # we do not want to raise RuntimeError(...) which would break an app on a Linux platform
        ...
```

marstr: I like this :) Something I wouldn't have thought of, but it definitely simplifies the consumption of this API. I'll have those changes in shortly

marstr: Fixed in a68ba24616da7d9d810a5bc33b2d3d0ff3817244

msal_extensions/token_cache.py
marstr: Fixed in a68ba24616da7d9d810a5bc33b2d3d0ff3817244

PR #16, opened 2019-05-17
Adding basic TokenProviders

### LINKED ISSUE #7: Create Token Provider Chain
If there are multiple ambient tokens present in the system, there should be a prioritized list of where to find them and which to choose.

The semantics of this fit well with the decorator/proxy pattern in general, and is how @devigned is implementing this in the .NET extensions repository here:

https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/13

### LINKED ISSUE #8: Create Service Principal Token Provider
Some folks will continue to authenticate using Service Principals (SPs). If there are available SP credentials via popular environment variables (i.e. `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET`), or if the MSAL Cache is extended to hold these credentials (see https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/14), there should be a token provider to retrieve them.

marstr:
* Define abstract class TokenProvider to define a contract for
implementations.
* Create ServicePrincipalProvider which uses well-known environment
variables should they be set.
* Create a "ProviderChain" which will allow a cascading search for
TokenProviders that can fulfill a request.

Closes #7 
Closes #8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -301,6 +301,7 @@ __pycache__/
 # Python Auxiliary Tools
 *.egg-info/
 .tox/
+.mypy_cache/
 
 # Cake - Uncomment if you are using it
 # tools/**
--- a/msal_extensions/__init__.py
+++ b/msal_extensions/__init__.py
@@ -1,4 +1,5 @@
 """Provides auxiliary functionality to the `msal` package."""
 __version__ = "0.0.1"
 
-from .token_provider import TokenProvider, TokenProviderChain, ServicePrincipalProvider, DEFAULT_TOKEN_CHAIN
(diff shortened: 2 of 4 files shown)


marstr:
More context around this PR, the idea is to emulate the behavior that was added to this repository's .NET sister via this PR: https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/13

rayluo:
Closing old PR/issues. The "token provider chain" concept has since been implemented in Azure SDK's Azure Identity package. The MSAL EX package here specialized to become a lower-level building block that focuses on data encryption and cross-process locking.

PR #18, opened 2019-05-21
Add TravisCI

marstr:
Until Azure DevOps is setup, I'd like to have tests running somewhere!


msal_extensions/cache_lock.py
```diff
@@ -1,16 +1,18 @@
...
 import sys
 import errno
 import portalocker
 
 
 class CrossPlatLock(object):
-    """
-    Offers a mechanism for waiting until another process is finished interacting with a shared resource. This is
-    specifically written to interact with a class of the same name in the .NET extensions library.
+    """ Offers a mechanism for waiting until another process is finished interacting with a shared
```
rayluo: Even if you keep its original form, I would not be such nit-picking on a docsting of a non-public helper. :-)  But since you already went the extra mile to use pylint as your self-imposed constraint (thumbs up!), I think you might want to know the [canonical way to write such docstring](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings). This is just a FYI. You do not have to change it. I'm giving a ship-it to this PR now.

marstr: Fixed in b908617

PR #19, opened 2019-05-22
Adding default value for LOCALAPPDATA

marstr:
Now that the [GitHub webhook outage is done](https://www.traviscistatus.com/incidents/c9n15td9svp0), it exposed some CI failures that should have been caught in #14! This fixes the TypeError that is causing Windows jobs to fail.

### DIFF
--- a/msal_extensions/windows.py
+++ b/msal_extensions/windows.py
@@ -118,7 +118,7 @@ class WindowsTokenCache(msal.SerializableTokenCache):
     """
     def __init__(self,
                  cache_location=os.path.join(
-                     os.getenv('LOCALAPPDATA'),
+                     os.getenv('LOCALAPPDATA', ''),
                      '.IdentityService',
                      'msal.cache'),
                  entropy=''):
--- a/msal_extensions/windows.py
+++ b/msal_extensions/windows.py
@@ -84,7 +84,7 @@ class WindowsDataProtectionAgent(object):
                 _LOCAL_FREE(result.pbData)
 
         err_code = _GET_LAST_ERROR()
-        raise OSError(1, '', '', err_code)
+        raise OSError(256, '', '', err_code)
 
     def unprotect(self, cipher_text):
         # type: (bytes) -> str
@@ -113,7 +113,7 @@ class WindowsDataProtectionAgent(object):
             finally:
                 _LOCAL_FREE(result.pbData)
         err_code = _GET_LAST_ERROR()
-        raise OSError(1, '', '', err_code)
+        raise OSError(256, '', '', err_code)
 
 
 class WindowsTokenCache(msal.SerializableTokenCache):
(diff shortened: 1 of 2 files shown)


msal_extensions/windows.py
```diff
@@ -109,7 +109,7 @@ class WindowsTokenCache(msal.SerializableTokenCache):
     """
     def __init__(self,
                  cache_location=os.path.join(
-                     os.getenv('LOCALAPPDATA'),
+                     os.getenv('LOCALAPPDATA', ''),
```
rayluo: Interesting. With hindsight, env variable "LOCALAPPDATA" would probably not exist on Linux and macOS either? In such case, the default value would write into a file named ".IdentityService/msal.cache" in current app's cwd, which defeats the purpose of sharing token across apps possibly developed in different languages.

We could consider using something [like this](https://docs.python.org/3/library/os.path.html#os.path.expanduser). Please test it, and also bring this up to .Net extension folks.

```python
os.getenv('LOCALAPPDATA', os.path.expanduser('~'))
```

marstr: Fixed with the combination of 934ef7d and 3da5724

rayluo: The fix looks good. I'm closing this thread. Once you get the remaining Travis CI errors fixed, we would be able to wrap up this.

marstr:
So the last error is a doozy, and I believe to be isolated to the way TravisCI is running its Windows machines. I've double checked, and the tests work for me. So the way I'd like to handle it, if folks are okay with it: recognize that the test is about to fail, and then detect that we are in Travis CI. If we are, just skip it. However, I've added an issue (#21) to follow up on this.

msal_extensions/windows.py
```diff
@@ -81,7 +82,9 @@ def protect(self, message):
                 return result.raw()
             finally:
                 _LOCAL_FREE(result.pbData)
-        return b''
+
+        err_code = _GET_LAST_ERROR()
+        raise OSError(1, '', '', err_code)
```
rayluo: In one of your [earlier attempt](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/19/commits/332f0bcb2b90c27b72df738f4219a6acca77c70c), you used `None` as the first parameter. Would that work? If so, I would suggest to stick with `None`. If it has to be an integer, then maybe choose a [non-existing integer](http://minirighi.sourceforge.net/html/errno_8h-source.html) such as `127`?

After that, the [automatic translation](https://docs.python.org/3/library/exceptions.html#OSError.winerror) would kick in and translate it into a who-knows-what value, therefore the end result would probably still be the same, determined by that auto-translation mechanism (and the weather). But it would feel semantically more appropriate to NOT arbitrarily choose an potentially misleading (albeit admittedly close enough) error code `1`.

marstr: Fixed in 15469c3310e4cbf66827a54f68efdf411be35345

PR #20, opened 2019-05-22
Fixing all remaining pylint errors in windows.py

marstr:
This should fix the Linux and Mac CI runs. In tandem with #19, the goal is to fix the CI build on the `dev` branch.


marstr:
I've confirmed that all of the CI jobs that this PR seeks to fix have passed. The Windows jobs that failed are out-of-scope of this PR.

PR #23, opened 2019-05-24
Adopt msal~=0.4

### LINKED ISSUE #22: Adopt msal~=0.4
The existing functionality in this repository only targets https://pypi.org/project/msal/0.3.0/ and compatible.

Given the new release, 0.4.0, the functionality in this repository must be updated.

marstr:
Fixes #22

### DIFF
--- a/msal_extensions/windows.py
+++ b/msal_extensions/windows.py
@@ -165,7 +165,10 @@ class WindowsTokenCache(msal.SerializableTokenCache):
                 except IOError as exp:
                     if exp.errno != errno.ENOENT:
                         raise exp
-            super(WindowsTokenCache, self).modify(credential_type, old_entry, new_key_value_pairs=new_key_value_pairs)
+            super(WindowsTokenCache, self).modify(
+                credential_type,
+                old_entry,
+                new_key_value_pairs=new_key_value_pairs)
             self._write()
 
     def find(self, credential_type, **kwargs):  # pylint: disable=arguments-differ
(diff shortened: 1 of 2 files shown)


msal_extensions/windows.py
```diff
@@ -157,26 +157,15 @@ def add(self, event, **kwargs):
...
-    def remove_rt(self, rt_item):
-        with CrossPlatLock(self._lock_location):
-            if self._needs_refresh():
-                try:
-                    self._read()
-                except IOError as exp:
-                    if exp.errno != errno.ENOENT:
-                        raise exp
-            super(WindowsTokenCache, self).remove_rt(rt_item)
+            super(WindowsTokenCache, self).modify(credential_type, old_entry, new_key_value_pairs)
```
rayluo: Nitpick: The current implementation is funtionally correct. However, because that `new_key_value_pairs` is a [keyword argument](https://docs.python.org/2/tutorial/controlflow.html#keyword-arguments), we should better explicitly use it as so:

```python
    super(WindowsTokenCache, self).modify(credential_type, old_entry,
        new_key_value_pairs=new_key_value_pairs)
```
or again use the `**kwargs` pattern:
```python
def modify(self, credential_type, old_entry, **kwargs):
    ...
    super(...).modify(credential_type, old_entry, **kwargs)
```

That way, the kwargs would remain orderless, which is semantically good. Should this method in base class obtain yet another new keyword argument in the future but with different order, your code here won't be broken. (Well, upstream library should NOT introduce new keyword argument in different order because it would be a subtle breaking change, but in reality, it happens anyway - it happened once in MSAL Python and [bit me too](https://github.com/AzureAD/microsoft-authentication-library-for-python/commit/2d6c373c29384d07ee294cb3f9fdc2dfab7dd0f7#diff-9803a6c326f35bba15eaa6b33829a874R97). My bad.)

marstr: Doesn't sound like a nit to me! This is good feedback :) fixed in 9664d6b

PR #27, opened 2019-07-16
Removing `FileTokenCache.add` overload.

### LINKED ISSUE #26: Non re-entrant lock blocking extension use with msal~=0.4.1
Repro:

1. Clone this repository on MacOS or Windows
1. Create new virtual environment
1. Run pip install, ensure `pip list | grep msal` (or similar command outside of bash) is greater or equal to 0.4.1.
1. Run pytest

Expected:
- Tests will pass

Actual:
- Tests hang indefinitely.

What happened is that in [8a7495a006eafd75179c0975b7fc52fddf733ca5](https://github.com/AzureAD/microsoft-authentication-library-for-python/commit/8a7495a006eafd75179c0975b7fc52fddf733ca5) the `add` function got refactored to call `self.modify`. In doing so, because of the inheritance involved, it is calling our `FileTokenCache`'s implementation of these functions and causing us to try to take a second lock on a file we're already locked on. This is causing an indefinite hang.

Amusingly, this is the exact problem talked about here: https://www.youtube.com/watch?v=YXiaWtc0cgE

marstr:
Fixes #26

### DIFF
--- a/msal_extensions/token_cache.py
+++ b/msal_extensions/token_cache.py
@@ -77,18 +77,6 @@ class FileTokenCache(msal.SerializableTokenCache):
         with open(self._cache_location, 'r') as handle:
             return handle.read()
 
-    def add(self, event, **kwargs):
-        with CrossPlatLock(self._lock_location):
-            if self._needs_refresh():
-                try:
-                    self.deserialize(self._read())
-                except IOError as exp:
-                    if exp.errno != errno.ENOENT:
-                        raise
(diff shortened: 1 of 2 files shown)


marstr:
Pro: Tests-pass
Con: The lock is now scoped so that it isn't for the duration of this library's add operation.

I'm not sure how much of a con that really is, as at the end of the day the only trouble would be with cache merging semantics. I'm open to your interpretation of that!

PR #32, opened 2019-09-26
Move required MSAL version to <1.0.0

rlrossiter:


setup.py
```diff
@@ -16,7 +16,7 @@
         'Development Status :: 2 - Pre-Alpha',
     ],
     install_requires=[
-        'msal>=0.4.1,<0.6.0',
+        'msal<1.0.0',
```
rayluo: It is OK to extend the upper bound to "<1.0.0", but we don't normally need to relax the lower bound. Such a specific lower bound was introduced for a reason (I can share more info on that historical reason, upon request :-) ).

PR #38, opened 2019-11-01
MSAL Extensions for Python 0.1.3

rayluo:
Release Notes:

* Provides same functionality as the previous version
* Can work with MSAL Python 1.0.0 or higher, until next MSAL Python major version bump
* Include license file in build artifacts (#37)


abhidnya13:
Hey @rayluo 
I think we need to add a setup.cfg to this repo as well ?

PR #41, opened 2019-11-11
Skip pywin32 226

rayluo:
Release 0.1.4

* This is [a temporary workaround which would otherwise be needed in our downstream package Azure SDK for Python](https://github.com/Azure/azure-sdk-for-python/issues/8573).

CC: @chlowell

### DIFF
--- a/msal_extensions/__init__.py
+++ b/msal_extensions/__init__.py
@@ -1,5 +1,5 @@
 """Provides auxiliary functionality to the `msal` package."""
-__version__ = "0.1.3"
+__version__ = "0.1.4"
 
 import sys
 
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,8 @@ setup(
     install_requires=[
         'msal>=0.4.1,<2.0.0',
         'portalocker~=1.0',
+        # Skip pywin32 226 release. See https://github.com/mhammond/pywin32/issues/1439
+        'pywin32<=225; platform_system == "Windows"',
     ],
     tests_require=['pytest'],
 )


lmazuel:
@rayluo that should be a short term fix only, because you are forcing customers to freeze this dep in time. If they use it in their own application for something else, they might be broken with a "dependency hell" at some point.

Could we agree that long term fix should be to remove that? Or at least consider this as an option if thye are not fixing this fast enough?

rayluo:
@lmazuel Thanks for calling this out! We did mention "temporary" in the proposed release note in this PR's description. And yes, once the upstream pywin32 fixes this from their end, we would then have to issue yet another patch to revert this change.

To be more clear, we wish we would NOT have to cut such a temporary release in the first place, to work around something that is not even in the scope of this repo/project.
We do it here, for Azure SDK for Python team, because this repo is more close to the upstream issue than Azure SDK for Python is. :-)

rayluo:
FYI: Our [upstream package `portalocker` has kindly adopted a similar change](https://github.com/WoLpH/portalocker/issues/48#event-2789615558) (with a reasonable assumption that the next release of pywin32 would fix it). Credit goes to @wolph

Stay tuned.

rayluo:
@chlowell @lmazuel:
Thanks to its author @wolph, our upstream package [portalocker 1.5.2 has just been released to PyPI, which excludes the problematic pywin32 226](https://github.com/WoLpH/portalocker/issues/48#issuecomment-552678237).

We no longer need this PR here.

PR #43, opened 2020-03-27
Adding explicit unlocking using portalocker

### LINKED ISSUE #42: The cache lock cannot be removed on Windows 10 if placed inside the user home directory.
I'm trying to create a TokenCache on Windows 10 and store a cache at 
```
~/.myapp/token.cache
```
As a result the lockfile is created at:
```
~/.myapp/token.cache.lockfile
```
However this lockfile is not removed in `CrossPlatLock.__exit__` and it is completely blocked by the  operationg system (no visible process) untill the further reboot. It looks like the file is not yet unlocked when being removed (I may be wrong) creating some kind of a hanging file handle. 

Here is a simple way to reproduce:
```
path = os.path.expanduser('~/.myapp/test.lockfile')
with CrossPlatLock(path):
    print('hi')
# One more time
with CrossPlatLock(path):
    print('hi')
    
<ipython-input-117-002d28c082a3> in <module>                                           
----> 1 with CrossPlatLock(path):                                                      
      2     print('hi')                                                                
      3                                                                                
                                                                                       
<ipython-input-116-89428a9dd969> in __enter__(self)                                    
     11         pid = os.getpid()                                                      
     12                                                                                
---> 13         self._fh = open(self._lockpath, 'wb+', buffering=0)                    
     14         portalocker.lock(self._fh, portalocker.LOCK_EX)                        
     15         self._fh.write('{} {}'.format(pid, sys.argv[0]).encode('utf-8'))       
                                                                                       
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\XXXXYYY/.myapp/test.lockfile'
```
Explicitly unlocking the file before closing it solves the issue:
```
portalocker.unlock(self._fh)
```
I know that portalocker claims that the explicit unlock is not necessary, but would it be possible to add it to prevent the problem?

abhidnya13:
Resolves #42

### DIFF
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -16,9 +16,9 @@ class CrossPlatLock(object):
                                       flags=portalocker.LOCK_EX, buffering=0)
 
     def __enter__(self):
-        fh = self._lock.__enter__()
-        fh.write('{} {}'.format(os.getpid(), sys.argv[0]).encode('utf-8'))
-        return fh
+        file_handle = self._lock.__enter__()
+        file_handle.write('{} {}'.format(os.getpid(), sys.argv[0]).encode('utf-8'))
+        return file_handle
 
     def __exit__(self, *args):
         self._lock.__exit__(*args)
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -12,18 +12,16 @@ class CrossPlatLock(object):
     """
     def __init__(self, lockfile_path):
         self._lockpath = lockfile_path
-        self._fh = None
+        self._lock = portalocker.Lock(lockfile_path, mode='wb+',
+                                      flags=portalocker.LOCK_EX, buffering=0)
 
     def __enter__(self):
-        pid = os.getpid()
-
-        self._fh = open(self._lockpath, 'wb+', buffering=0)
-        portalocker.lock(self._fh, portalocker.LOCK_EX)
-        self._fh.write('{} {}'.format(pid, sys.argv[0]).encode('utf-8'))
+        fh = self._lock.__enter__()
+        fh.write('{} {}'.format(os.getpid(), sys.argv[0]).encode('utf-8'))
+        return fh
 
     def __exit__(self, *args):
-        portalocker.unlock(self._fh)
-        self._fh.close()
+        self._lock.__exit__(*args)
         try:
             # Attempt to delete the lockfile. In either of the failure cases enumerated below, it is
             # likely that another process has raced this one and ended up clearing or locking the
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -17,7 +17,8 @@ class CrossPlatLock(object):
             mode='wb+',
             # In posix systems, we HAVE to use LOCK_EX(exclusive lock) bitwise ORed
             # with LOCK_NB(non-blocking) to avoid blocking on lock acquisition.
-            # More information here: https://docs.python.org/3/library/fcntl.html#fcntl.lockf
+            # More information here:
+            # https://docs.python.org/3/library/fcntl.html#fcntl.lockf
             flags=portalocker.LOCK_EX | portalocker.LOCK_NB,
             buffering=0)
 
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -12,8 +12,14 @@ class CrossPlatLock(object):
     """
     def __init__(self, lockfile_path):
         self._lockpath = lockfile_path
-        self._lock = portalocker.Lock(lockfile_path, mode='wb+',
-                                      flags=portalocker.LOCK_EX, buffering=0)
+        self._lock = portalocker.Lock(
+            lockfile_path,
+            mode='wb+',
+            # In posix systems, we HAVE to use LOCK_EX(exclusive lock) bitwise ORed
+            # with LOCK_NB(non-blocking) to avoid blocking on lock acquisition.
+            # More information on this here: https://docs.python.org/3/library/fcntl.html#fcntl.lockf
+            flags=portalocker.LOCK_EX | portalocker.LOCK_NB,
+            buffering=0)
 
     def __enter__(self):
         file_handle = self._lock.__enter__()
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -22,6 +22,7 @@ class CrossPlatLock(object):
         self._fh.write('{} {}'.format(pid, sys.argv[0]).encode('utf-8'))
 
     def __exit__(self, *args):
+        portalocker.unlock(self._fh)
         self._fh.close()
         try:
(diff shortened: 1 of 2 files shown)


msal_extensions/cache_lock.py
```diff
@@ -22,6 +22,7 @@ def __enter__(self):
         self._fh.write('{} {}'.format(pid, sys.argv[0]).encode('utf-8'))
 
     def __exit__(self, *args):
+        portalocker.unlock(self._fh)
```
henrik-me: >portalocker [](start = 8, length = 11)

isn't there also a newer version of portalocker available?

rayluo: The change here is functionally correct. This can be our fall back option.

As we briefly mentioned in yesterday's talk, let's also try to see if we can build our `CrossPlatLock` on top of `portalocker.Lock`. Something like this. (Note that this change touches much more lines than our fall-back option, so we should test it even more carefully. But the benefit is we might be able to separate the behaviors in a clean way, once and for all. If it works, we would want to use it.)

```python
def __init__(self, lockfile_path):
    self._lockpath = lockfile_path
    self._lock = portalocker.Lock(lockfile_path, ...)

def __enter__(self):
    fh = self._lock.__enter__()
    fh.write('{} {}'.format(os.getpid(), sys.argv[0]).encode('utf-8'))
    return fh

def __exit__(self, *args):
    self._lock.__exit__(*args)
    try:
        # Attempt to delete the lockfile. In either of the failure cases enumerated below, it is
        # likely that another process has raced this one and ended up clearing or locking the
        # file for itself.
        os.remove(self._lockpath)
    except OSError as ex:
        if ex.errno != errno.ENOENT and ex.errno != errno.EACCES:
            raise
```

msal_extensions/cache_lock.py
```diff
@@ -22,6 +22,7 @@ def __enter__(self):
...
         self._fh.close()
         try:
             # Attempt to delete the lockfile. In either of the failure cases enumerated below, it is
```
rayluo, line 28: I'm quoting @bgavrilMS 's [comment from another PR](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/44#issuecomment-610326593), and answer it here, because the locking functionality is actually provided here.

> ... it is fairly important to delete the file if the process which created it crashes. Otherwise, you will block all other consumers and the only "way out" is for the user to manually delete the lock file. 

> Should not be difficult to implement, in Java this is done like [this](https://github.com/AzureAD/microsoft-authentication-extensions-for-java/blob/master/src/main/java/com/microsoft/aad/msal4jextensions/CrossProcessCacheFileLock.java#L46) and in .net like [this](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/bac27e967509ee0c37037f2d4ebf9cd51d31d168/src/Shared/CrossPlatLock.cs#L52)

That "DeleteOnClose" behavior provided by in C# and Java is indeed nice. I kind of remember that there was an equivalent thing in the low-level C API (but I [could not find it now](https://www.gnu.org/software/libc/manual/html_node/Opening-and-Closing-Files.html)). Regardless, it is unfortunately that [Python's built-in open() function](https://docs.python.org/3/library/functions.html#open) does not provide this functionality. So the next best thing we can do, is to explicitly remove it at the line below, and live with the potential exception.

bgavrilMS: I suppose a workaround for not having this functionality is to use good old "try" / "finally" in the objects that use the lock. Similar to the "Disposable" concept in C#. I believe you already do that.

bgavrilMS: Check out this SO question, the answer with ~21 votes shows how to get the "DeleteOnClose" functionality https://stackoverflow.com/questions/54772244/whats-a-good-way-to-detect-when-a-python-program-exits-or-crashes

It seems to be different on Win / Linux

rayluo: But that SO q&a were about locking, which is a topic that we happened to already choose its ~39 votes answer for (and I personally even tend to agree with that answer's second part "use a database if at all possible"). :-)

Very few (if any at all) locking library/conversation out there would cover file deletion behavior, simply because file deletion is not part of the usual locking mechanism, which in turn is based on OS's `fcntl.LOCK_EX` on a file handler, not based on application's detection of the existence of a file. It is just our MSAL EX protocol happened to choose to do file deletion (but why?), so I guess we have to implement that by ourselves.

All in all, using an explicit `os.remove(...)` and handling its exception properly, seems adequate.

abhidnya13:
https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/Default-timeout-and-retry-settings-for-locking-mechanisms

msal_extensions/cache_lock.py
```diff
@@ -12,17 +12,22 @@ class CrossPlatLock(object):
     """
     def __init__(self, lockfile_path):
         self._lockpath = lockfile_path
-        self._fh = None
+        self._lock = portalocker.Lock(
+            lockfile_path,
+            mode='wb+',
+            # In posix systems, we HAVE to use LOCK_EX(exclusive lock) bitwise ORed
+            # with LOCK_NB(non-blocking) to avoid blocking on lock acquisition.
+            # More information here: https://docs.python.org/3/library/fcntl.html#fcntl.lockf
```
rayluo, line 20: Nit: Given that a URL tends to be long, it is usually better to put it in its own line to help keep lines within the 80-character long recommendation by PEP 8. I'm still going to give an approval on this PR. :-)

```suggestion
            # More information here:
            # https://docs.python.org/3/library/fcntl.html#fcntl.lockf
```

bluca:
@abhidnya13 @rayluo just to be sure, what happens on Linux if portalocker doesn't have the fix? Will something break, or will it be an issue only in Windows?

rayluo:
@bluca According to #42, the specific issue, i.e. lock file is not being deleted, is only reported in Windows.

Did you observe any issue on Linux? If so, what is the symptom? Please create a new issue to describe it.

bluca:
> @bluca According to #42, the specific issue, i.e. lock file is not being deleted, is only reported in Windows.
> 
> Did you observe any issue on Linux? If so, what is the symptom? Please create a new issue to describe it.

I did not see any issue - I am wondering whether I need to enforce the dependency on Linux or not when uploading the new version to Debian/Ubuntu.

rayluo:
@bluca If you are using this [MSAL Extensions for Python library, it already enforces that dependency](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/0.2.0/setup.py#L21) (on all platforms, for that matter). If you are using portalocker directly, you can make that decision. You may even open an issue in that repo and hear the input from its author, who is very responsive to github issues.

bluca:
> @bluca If you are using this [MSAL Extensions for Python library, it already enforces that dependency](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/0.2.0/setup.py#L21) (on all platforms, for that matter).

No, it doesn't - unless I make it. Hence the question - should it? Is it really necessary for Linux?

PR #44, opened 2020-04-02
Encryption on Linux based on libsecret

### LINKED ISSUE #4: Create a GNOME Keyring Token Cache (Linux only)
Linux has a few solutions for protecting secrets. For people using the Desktop Environment(DE) [GNOME](https://www.gnome.org), [GNOME  Keyring](https://wiki.gnome.org/Projects/GnomeKeyring) is the natural choice. While it is not as ubiquitous as DP API for Windows users or KeyChain for MacOS users, it is a good way to get fairly significant coverage of Linux developers.

Just like #1 and #3, we should implement a class to facilitate interactions with GNOME Keyring that extends [SerializableTokenCache](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/f76f3c37c6bf7747971947d300f0a50d3ec701f7/msal/token_cache.py#L204). However, unlike those platforms, as of time of writing there is no .NET support for this secret backend.

Researching a little this afternoon, it seems like it is not unprecedented for folks who are not using GNOME as their DE to take a dependency on just this package. This may prove useful for us when running in containers, or other Linux environments that have no DE whatsoever. Experimentally, I was able to install `gnome-keyring` with only standard repositories in both an ubuntu:disco and centos7 container.

rayluo:
Description
----------------
* ~The CI tests are currently failing, but only because some styling error. I'll fix it soon. This does NOT block your review.~ Fixed.
* So far I run pre-existing `test_agnostic_backend.py` to test it manually on Linux Desktop, and "it seems to work"<sup>TM</sup>
* I have ~NOT yet~ tested it in SSH console. Encryption is NOT available in an SSH console, and this exception will be raised, as expected: `gi.repository.GLib.GError: g-io-error-quark: Cannot autolaunch D-Bus without X11 $DISPLAY (0)`
* ~I'll~ add more API docs ~later~ below.
* This PR will fix #4

Usage
---------

1. The main interface of MSAL Extension for Python (MSAL EX Python) has historically been some platform-dependent `TokenCache` classes, whose instances can then be fed into MSAL Python's `PublicClientApplication("my_client_id", ..., token_cache=some_cache, ...)`. In this PR, we introduce one new `PersistedTokenCache(...)` which would then accept a platform-dependent persistence instance. (Azure CLI team may not need to care about this, if they would not directly consume MSAL Extension. Azure SDK team will use this.) Old API will be deprecated.

   * [Sample: Obtain a TokenCache suitable for your current OS](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/44/files#diff-eeb0aeb17a30896a2e0a290faa0b869c).

   *TokenCache includes a file lock, and auto-reload behavior under the hood*.

2. Since this PR, there will also be a lower level persistence layer being exposed. However, that the *persistence layer does NOT have a built-in lock*. You can use yet another lower level helper `CrossPlatLock` in MSAL EX Python for that purpose. We will explain that reason later.

   * [Sample 2: The persistence layer which separates from the token cache](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/44/files#diff-ea278a6fabad90ffa71c8628d07d2d39).

Q&A
---

Q: What is the relationship among Token Cache, Service Principle Storage, CrossPlatLock and Persistence (with Encryption)?

A: Token Cache was developed on top of CrossPlatLock and Persistence (with Encryption). Service Principle Storage would also want to reuse the same two underlying technology. But other than that, Token Cache and Service Principle Storage are two sets of different data with different characteristic.

* By definition, cache can tolerate data lose. When cache is lost, it is simply treated as cache miss, and the application would still work (and re-populate the cache soon). But credentials in Service Principle Storage might be irreplaceable.
* Token cache data is sensitive but not as sensitive as password. MSAL never stores passwords in token cache. Service Principle Storage would store passwords. So it is probably better to store them separately and could apply extra protection for service principles.
* Usage pattern might also be different. Token cache would be accessed very frequently during each app's run-time, thus can be very busy due to lots of lock/unlock operation. Service Principle Storage is likely needed only a few times, per an app's run-time. A separated physical file would mean less racing and better performance for each other.
* Locking strategy should be determined by application usage pattern, thus they are likely different between token cache and service principle storage. MSAL EX Python's token cache is designed for MSAL Python, lock is acquired/released based on cache operation, not based on low level file I/O. Similarly, service principle storage could acquire/release lock at a higher level, that's why the Persistence layer does not have built-in lock behavior.
* Similarly, Persistence layer does not provide file-changed detection, thus no auto-reload mechanism. That part has to be implemented in a real ServicePrincipleStorage class, which is outside of the scope of this Extensions.

Separated by 2 physically isolated files, the Token Cache and Service Principle Storage would look like 2 parallel stacks.

| Token Cache | Service Principles Storage (Not implemented here) |
| ------------------- | ------------------------ |
| CrossPlatLock | CrossPlatLock |
| Persistence (with encryption) | Persistence (with encryption) |

Known Behavior
-----------------------

* This Encryption on Linux implementation relies on a 3rd-party Python library `PyGObject` and all its dependencies.

* Imagine this sequence:

1. An end user Neo wakes up, logs in to his Linux Desktop, and uses MSAL to acquire tokens. This would work, and by default, the tokens would be stored encrypted.
2. In the morning, Neo leaves his home for work, and remotely connects back to his Linux computer via an ssh session. MSAL would NOT be able to access tokens encrypted in step 1, but app can opt in to let MSAL establish and maintain another (temporary - more on this later) empty token cache in a plain text file, as a fallback bahavior. Throughout the daytime, in any ssh session, MSAL will maintain tokens in such file.
3. At afternoon, Neo returns home, and logs in to his Linux Desktop again. MSAL will access his tokens stored from step 1, but not those from step 2. And, as an unobvious by-product, MSAL will wipe out the "temporary token cache" established at step 2.
4. At night, Neo heads out to a bar. He goes back to step 2 and start afresh.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,20 +6,35 @@ matrix:
     - python: "2.7"
       env: TOXENV=py27 PYPI=true
       os: linux
+      before_install:
+        - sudo apt update
+        - sudo apt install python-dev libgirepository1.0-dev libcairo2-dev gir1.2-secret-1
     - python: "3.5"
       env: TOXENV=py35
       os: linux
+      before_install:
+        - sudo apt update
+        - sudo apt install python3-dev libgirepository1.0-dev libcairo2-dev gir1.2-secret-1
     - python: "3.6"
       env: TOXENV=py36
       os: linux
+      before_install:
+        - sudo apt update
+        - sudo apt install python3-dev libgirepository1.0-dev libcairo2-dev gir1.2-secret-1
     - python: "3.7"
       env: TOXENV=py37
       os: linux
       dist: xenial
+      before_install:
+        - sudo apt update
+        - sudo apt install python3-dev libgirepository1.0-dev libcairo2-dev gir1.2-secret-1
     - python: "3.8"
       env: TOXENV=py38
       os: linux
       dist: xenial
+      before_install:
+        - sudo apt update
+        - sudo apt install python3-dev libgirepository1.0-dev libcairo2-dev gir1.2-secret-1
     - name: "Python 3.7 on macOS"
       env: TOXENV=py37
       os: osx
@@ -46,7 +61,8 @@ install:
   - pip install .
 
 script:
-  - pylint msal_extensions
+  - # Difficult to get .pylintrc working on both Python 2 & 3, and we don't have to
+  - if [ "$TOXENV" = "py37"]; then pylint msal_extensions; fi
   - tox
 
 deploy:
--- a/msal_extensions/libsecret.py
+++ b/msal_extensions/libsecret.py
@@ -1,0 +1,120 @@
+"""Implements a Linux specific TokenCache, and provides auxiliary helper types.
+
+This module depends on PyGObject. But `pip install pygobject` would typically fail,
+until you install its dependencies first. For example, on a Debian Linux, you need::
+
+    sudo apt install libgirepository1.0-dev libcairo2-dev python3-dev gir1.2-secret-1
+    pip install pygobject
+
+Alternatively, you could skip Cairo & PyCairo, but you still need to do all these
+(derived from https://gitlab.gnome.org/GNOME/pygobject/-/issues/395)::
+
+    sudo apt install libgirepository1.0-dev python3-dev gir1.2-secret-1
+    pip install wheel
+    PYGOBJECT_WITHOUT_PYCAIRO=1 pip install --no-build-isolation pygobject
+"""
+import logging
+
+import gi  # https://pygobject.readthedocs.io/en/latest/getting_started.html
+
+# pylint: disable=no-name-in-module
+gi.require_version("Secret", "1")  # Would require a package gir1.2-secret-1
+# pylint: disable=wrong-import-position
+from gi.repository import Secret  # Would require a package gir1.2-secret-1
+
+
+logger = logging.getLogger(__name__)
+
+class LibSecretAgent(object):
+    """A loader/saver built on top of low-level libsecret"""
+    # Inspired by https://developer.gnome.org/libsecret/unstable/py-examples.html
+    def __init__(  # pylint: disable=too-many-arguments
+            self,
+            schema_name,
+            attributes,  # {"name": "value", ...}
+            label="",  # Helpful when visualizing secrets by other viewers
+            attribute_types=None,  # {name: SchemaAttributeType, ...}
+            collection=None,  # None means default collection
+            ):  # pylint: disable=bad-continuation
+        """This agent is built on top of lower level libsecret API.
+
+        Content stored via libsecret is associated with a bunch of attributes.
+
+        :param string schema_name:
+            Attributes would conceptually follow an existing schema.
+            But this class will do it in the other way around,
+            by automatically deriving a schema based on your attributes.
+            However, you will still need to provide a schema_name.
+            load() and save() will only operate on data with matching schema_name.
+
+        :param dict attributes:
+            Attributes are key-value pairs, represented as a Python dict here.
+            They will be used to filter content during load() and save().
+            Their arbitrary keys are strings.
+            Their arbitrary values can MEAN strings, integers and booleans,
+            but are always represented as strings, according to upstream sample:
+            https://developer.gnome.org/libsecret/0.18/py-store-example.html
+
+        :param string label:
+            It will not be used during data lookup and filtering.
+            It is only helpful when/if you visualize secrets by other viewers.
+
+        :param dict attribute_types:
+            Each key is the name of your each attribute.
+            The corresponding value will be one of the following three:
+
+            * Secret.SchemaAttributeType.STRING
+            * Secret.SchemaAttributeType.INTEGER
+            * Secret.SchemaAttributeType.BOOLEAN
+
+            But if all your attributes are Secret.SchemaAttributeType.STRING,
+            you do not need to provide this types definition at all.
+
+        :param collection:
+            The default value `None` means default collection.
+        """
+        self._collection = collection
+        self._attributes = attributes or {}
+        self._label = label
+        self._schema = Secret.Schema.new(schema_name, Secret.SchemaFlags.NONE, {
+            k: (attribute_types or {}).get(k, Secret.SchemaAttributeType.STRING)
+            for k in self._attributes})
+
+    def save(self, data):
+        """Store data. Returns a boolean of whether operation was successful."""
+        return Secret.password_store_sync(
+            self._schema, self._attributes, self._collection, self._label,
+            data, None)
+
+    def load(self):
+        """Load a password in the secret service, return None when found nothing"""
+        return Secret.password_lookup_sync(self._schema, self._attributes, None)
+
+    def clear(self):
+        """Returns a boolean of whether any passwords were removed"""
+        return Secret.password_clear_sync(self._schema, self._attributes, None)
+
+
+def trial_run():
+    """This trial run will raise an exception if libsecret is not functioning.
+
+    Even after you installed all the dependencies so that your script can start,
+    or even if your previous run was successful, your script could fail next time,
+    for example when it will be running inside a headless SSH session.
+
+    You do not have to do trial_run. The exception would also be raised by save().
+    """
+    try:
+        agent = LibSecretAgent("Test Schema", {"attr1": "foo", "attr2": "bar"})
+        payload = "Test Data"
+        agent.save(payload)  # It would fail when running inside an SSH session
+        assert agent.load() == payload  # This line is probably not reachable
+        agent.clear()
+    except (gi.repository.GLib.Error, AssertionError):
+        message = (
+            "libsecret did not perform properly. Please refer to "
+            "https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/Encryption-on-Linux")  # pylint: disable=line-too-long
+        logger.exception(message)  # This log contains trace stack for debugging
+        logger.warning(message)  # This is visible by default
+        raise
+
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -1,0 +1,201 @@
+"""A generic persistence layer, optionally encrypted on Windows, OSX, and Linux.
+
+Should a certain encryption is unavailable, exception will be raised at run-time,
+rather than at import time.
+
+By successfully creating and using a certain persistence object,
+app developer would naturally know whether the data are protected by encryption.
+"""
+import abc
+import os
+import errno
+try:
+    from pathlib import Path  # Built-in in Python 3
+except:
+    from pathlib2 import Path  # An extra lib for Python 2
+
+
+try:
+    ABC = abc.ABC
+except AttributeError:  # Python 2.7, abc exists, but not ABC
+    ABC = abc.ABCMeta("ABC", (object,), {"__slots__": ()})  # type: ignore
+
+
+def _mkdir_p(path):
+    """Creates a directory, and any necessary parents.
+
+    This implementation based on a Stack Overflow question that can be found here:
+    https://stackoverflow.com/questions/600268/mkdir-p-functionality-in-python
+
+    If the path provided is an existing file, this function raises an exception.
+    :param path: The directory name that should be created.
+    """
+    if not path:
+        return  # NO-OP
+    try:
+        os.makedirs(path)
+    except OSError as exp:
+        if exp.errno == errno.EEXIST and os.path.isdir(path):
+            pass
+        else:
+            raise
+
+
+class BasePersistence(ABC):
+    """An abstract persistence defining the common interface of this family"""
+
+    is_encrypted = False  # Default to False. To be overridden by sub-classes.
+
+    @abc.abstractmethod
+    def save(self, content):
+        # type: (str) -> None
+        """Save the content into this persistence"""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def load(self):
+        # type: () -> str
+        """Load content from this persistence"""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def time_last_modified(self):
+        """Get the last time when this persistence has been modified"""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_location(self):
+        """Return the file path which this persistence stores (meta)data into"""
+        raise NotImplementedError
+
+
+class FilePersistence(BasePersistence):
+    """A generic persistence, storing data in a plain-text file"""
+
+    def __init__(self, location):
+        if not location:
+            raise ValueError("Requires a file path")
+        self._location = os.path.expanduser(location)
+        _mkdir_p(os.path.dirname(self._location))
+
+    def save(self, content):
+        # type: (str) -> None
+        """Save the content into this persistence"""
+        with open(self._location, 'w+') as handle:
+            handle.write(content)
+
+    def load(self):
+        # type: () -> str
+        """Load content from this persistence"""
+        with open(self._location, 'r') as handle:
+            return handle.read()
+
+    def time_last_modified(self):
+        return os.path.getmtime(self._location)
+
+    def touch(self):
+        """To touch this file-based persistence without writing content into it"""
+        Path(self._location).touch()  # For os.path.getmtime() to work
+
+    def get_location(self):
+        return self._location
+
+
+class FilePersistenceWithDataProtection(FilePersistence):
+    """A generic persistence with data stored in a file,
+    protected by Win32 encryption APIs on Windows"""
+    is_encrypted = True
+
+    def __init__(self, location, entropy=''):
+        """Initialization could fail due to unsatisfied dependency"""
+        # pylint: disable=import-outside-toplevel
+        from .windows import WindowsDataProtectionAgent
+        self._dp_agent = WindowsDataProtectionAgent(entropy=entropy)
+        super(FilePersistenceWithDataProtection, self).__init__(location)
+
+    def save(self, content):
+        super(FilePersistenceWithDataProtection, self).save(
+            self._dp_agent.protect(content))
+
+    def load(self):
+        return self._dp_agent.unprotect(
+            super(FilePersistenceWithDataProtection, self).load())
+
+
+class KeychainPersistence(BasePersistence):
+    """A generic persistence with data stored in,
+    and protected by native Keychain libraries on OSX"""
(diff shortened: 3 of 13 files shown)


.travis.yml
```diff
@@ -6,20 +6,35 @@ matrix:
     - python: "2.7"
       env: TOXENV=py27 PYPI=true
       os: linux
+      before_install:
+        - sudo apt update
+        - sudo apt install python-dev libgirepository1.0-dev libcairo2-dev gir1.2-secret-1
```
bgavrilMS, line 11: Very cool, do your tests actually run on a cloud machine with Linux? A bit surprised as I thought most distros require an X11 (e.g. Gnome Desktop) to be installed.

msal_extensions/libsecret.py
```diff
@@ -0,0 +1,64 @@
...
+        """Load a password in the secret service, return None when found nothing"""
+        return Secret.password_lookup_sync(self._schema, self._attributes, None)
+
+    def clear(self):
+        """Returns a boolean of whether any passwords were removed"""
+        return Secret.password_clear_sync(self._schema, self._attributes, None)
+
+def verify():
+    """Exception will be raised if libsecret is not functioning"""
+    agent = LibSecretAgent("Schema_Name Testing", {"attr1": "foo", "attr2": "bar"})
```
bgavrilMS, line 56: Nit: use some constants like "persistence_test"

msal_extensions/libsecret.py
```diff
@@ -0,0 +1,64 @@
...
+    message = "Test Data"
+    agent.save(message)  # This line would fail when running inside an SSH session
+    if agent.load() != message:
+        raise RuntimeError("libsecret did not perform properly")
```
bgavrilMS, line 60: Nit: include more details in the exception message, perhaps an aka.ms link

msal_extensions/token_cache.py
```diff
@@ -1,156 +1,119 @@
...
-        Inspects the file holding the encrypted TokenCache to see if a read is necessary.
-        :return: True if there are changes not reflected in memory, False otherwise.
-        """
+    def _reload_if_necessary(self):
+        # type: () -> None
+        """Reload cache from persistence layer, if necessary"""
         try:
-            updated = os.path.getmtime(self._cache_location)
-            return self._last_sync < updated
+            if self._last_sync < self._persistence.getmtime():
```
bgavrilMS, line 35: I think this needs more testing. I believe the .NET implementers found that timestamps are not granular enough to ensure locking.

msal_extensions/token_cache.py
```diff
@@ -1,156 +1,119 @@
...
-    def __init__(self,
-                 cache_location,
-                 lock_location=None):
-        super(FileTokenCache, self).__init__()
-        self._cache_location = cache_location
-        self._lock_location = lock_location or self._cache_location + '.lockfile'
+    def __init__(self, persistence, lock_location=None):
+        super(PersistedTokenCache, self).__init__()
+        self._lock_location = (
+            os.path.expanduser(lock_location) if lock_location
```
bgavrilMS, line 24: Just to confirm - the location of the lockfile is the same as the location of the cache file (or dummy cache file)..

bgavrilMS:
Generally looks good, I can only see one problem - the file lock semantics seem to be different? In .net and java, we open a file stream for writing as a method of "locking". The file is closed if the process dies. I haven't seen this here, but possibly I don't understand python code well enough?

A few more observations:

- You should be able to automate some E2E tests on those persistence classes. You can configure your CI to run tests on Win, Mac and Linux (not sure about Travis, but ADO definitely allows you do this). 

- It would be good to have an a simple app that demonstrates usage. This helped me drive a lot of the conversations with customers. It would also showcase how to use the Linux fallback option.

msal_extensions/libsecret.py
rayluo: IMHO, unless a string would be subjected to future localization need (which it isn't in our SDK scenarios), the pattern of "use pre-defined constants rather than repeated on-the-spot magic strings" does not give a real benefit, when that magic string is used once and only once.

msal_extensions/libsecret.py
rayluo: Good idea. Speaking about aka.ms link, I don't currently know how to set one up. Have you already created one? Let's follow up offline on this.
By the way, after testing during weekend (and documenting my findings at line 58), I actually realized that this line 60 will NOT be reachable, because another exception would be raised at line 58. It would be better to refactor this part of code so that the aka.ms link would be available from line 58 too.

msal_extensions/token_cache.py
rayluo: By "same location", did you mean "same folder/directory"? Indeed, by default we just add a ".lockfile" suffix to the cache filename, to form the lock filename (see line 25 below). Historically the API surface also accepts a customized lockfile location, and perhaps that is a feature also available in .Net/Java too, although I doubt how many people actually use that behavior.

msal_extensions/token_cache.py
rayluo: Here the timestamp is only used to determine whether the data copy in memory becomes stale (and it seems we won't have other ways to detect that).

This timestamp check is NOT used for locking mechanism. The lock has already been acquired when the app reaches this line of code.

bgavrilMS:
Yes, the locking mechanism sounds fine - if the file is present, it means the cache is locked.

But it is fairly important to delete the file if the process which created it crashes. Otherwise, you will block all other consumers and the only "way out" is for the user to manually delete the lock file. 

Should not be difficult to implement, in Java this is done like [this](https://github.com/AzureAD/microsoft-authentication-extensions-for-java/blob/master/src/main/java/com/microsoft/aad/msal4jextensions/CrossProcessCacheFileLock.java#L46) and in .net like [this](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/bac27e967509ee0c37037f2d4ebf9cd51d31d168/src/Shared/CrossPlatLock.cs#L52)

msal_extensions/libsecret.py
rayluo: I ended up providing a url in the error message, and that link points to a wiki page, for now. Resolving this conversation.

msal_extensions/token_cache.py
```diff
@@ -1,156 +1,113 @@
...
+        entropy='',  # For Windows
+        service_name="", account_name="",  # For OSX
+        schema_name="", attributes=None, fallback_to_plaintext=False,  # For Linux
+        **kwargs):  # Extra parameters will be relayed to LibsecretPersistence
+    """Build a token cache which uses native encryption mechanism on local platform.
 
-class OSXTokenCache(FileTokenCache):
-    """A SerializableTokenCache implementation which uses native Keychain libraries to protect your
-    tokens.
+    It will use DBAPI when running on Windows, Keychain when running on OSX,
```
abhidnya13, line 67: Noticed a typo for DPAPI while working on the 2in1 branch

msal_extensions/token_cache.py
```diff
@@ -1,156 +1,113 @@
...
+        warnings.warn("Use build_token_cache(...) instead", DeprecationWarning)
+        super(WindowsTokenCache, self).__init__(
+            FilePersistenceWithDataProtection(cache_location, entropy=entropy))
+
+
+class OSXTokenCache(PersistedTokenCache):
+    """A token cache which uses native Keychain libraries to encrypt your tokens."""
     def __init__(self,
                  cache_location,
                  service_name='Microsoft.Developer.IdentityService',
```
jianghaolu, line 107: This default value is not defined in the .NET or Java extension - though it is exactly where Visual Studio for Mac stores the token cache. I'm fine with having a default value, though just want to point it out in case anyone else has an opinion.

msal_extensions/token_cache.py
```diff
@@ -1,156 +1,113 @@
...
+    libsecret when running on Linux and optionally fallback to plain texxt file.
     """
-
+    persistence = build_persistence(
+        location,
+        entropy=entropy,
+        service_name=service_name,
+        account_name=account_name,
+        schema_name=schema_name,
+        attributes=attributes or {"msal_token_cache": "1.0"},
```
jianghaolu, line 76: Another default value not defined in .NET or Java. I'm not able to verify this though. Do you know where this value comes from?

msal_extensions/token_cache.py
rayluo: Good observation, @jianghaolu , even on something NOT being changed, thus not highlighted in this PR! This is indeed a good topic to raise. Here come my thoughts.

* The current default values were pre-existing before this PR. I was not sure where/how they came from. I suspect it matches some values other Microsoft-owned products chose to use. Thanks for your confirmation.
* However, I would actually prefer that all our MSAL EX to NOT provide such default value here. Because this MSAL EX will also be used by 3rd-party app developers. And 1st-party and 3rd-party apps might not want to share tokens with each other.
* In fact, this PR intends to deprecate this entire `OSXTokenCache` class. And the new `build_token_cache(...)` factory method uses a set of non-Microsoft settings as default values. All apps are expected to provide their own settings.

msal_extensions/token_cache.py
rayluo: Again, this is just some placeholder that I deliberately chose to be different. Perhaps we should even choose to NOT provide a default at all, and raise ValueError when app developer did not provide a specific value?

msal_extensions/token_cache.py
```diff
@@ -1,156 +1,113 @@
...
-        with open(self._cache_location, 'rb') as handle:
-            cipher_text = handle.read()
-        return self._dp_agent.unprotect(cipher_text)
 
+def build_token_cache(  # pylint: disable=too-many-arguments
+        location,
+        entropy='',  # For Windows
+        service_name=None, account_name=None,  # For OSX
+        schema_name="", attributes=None, fallback_to_plaintext=False,  # For Linux
+        **kwargs):  # Extra parameters will be relayed to LibsecretPersistence
```
chlowell, line 64: As a user, I'd prefer my own code to this helper because the union of options is too mysterious for me. It looks like sorting out how each persistence implementation uses which options is a prerequisite for calling this method anyway.

msal_extensions/token_cache.py
jianghaolu: It makes sense to keep it backward compatible in here and require full settings from the new configuration API.

msal_extensions/token_cache.py
jianghaolu: I couldn't find this value in either the Java or .NET repo. Though in my Java testing I'm able to get Keyring work without providing this attribute.

msal_extensions/libsecret.py
```diff
@@ -0,0 +1,78 @@
...
+        """Returns a boolean of whether any passwords were removed"""
+        return Secret.password_clear_sync(self._schema, self._attributes, None)
+
+def verify():
+    """Exception will be raised if libsecret is not functioning"""
+    try:
+        agent = LibSecretAgent("Test Schema", {"attr1": "foo", "attr2": "bar"})
+        payload = "Test Data"
+        agent.save(payload)  # It would fail when running inside an SSH session
+        assert agent.load() == payload  # This line is probably not reachable
```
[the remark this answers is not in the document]
rayluo: It shouldn't matter, as it is just dummy data. But OK, we will add a `clear()` line in next rebase.

msal_extensions/persistence.py
```diff
@@ -0,0 +1,209 @@
...
+            return LibsecretPersistence(
+                # By using same location as the fall back option below,
+                # this would override the unencrypted data stored by the
+                # fall back option.  It is probably OK, or even desired
+                # (in order to aggressively wipe out plain-text persisted data),
+                # unless there would (frequently) be a desktop session and
+                # a remote ssh session being active simultaneously.
+                location,
+                schema_name=schema_name,
+                attributes=attributes or {"attr": "value"},
```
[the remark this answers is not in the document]
rayluo: Yep, it has already been removed in a rebase earlier today. Now the caller need to provide a value. No more meaningless magic defaults. :-P

msal_extensions/persistence.py
```diff
@@ -0,0 +1,209 @@
...
+        return self._dp_agent.unprotect(
+            super(FilePersistenceWithDataProtection, self).load())
+
+
+class KeychainPersistence(BasePersistence):
+    """A generic persistence with data stored in,
+    and protected by native Keychain libraries on OSX"""
+    is_encrypted = True
+
+    def __init__(self, location, service_name, account_name, dummy=" "):
```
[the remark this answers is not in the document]
rayluo: Yeah it was not obvious. We should have documented it inside our docstring. The thing is, some of our downstream applications chose to monitor a FileChanged event on that `location`, and it is believed that only a real file I/O would trigger that FileChanged event. In particular, a `touch` action to change file metadata would not trigger that event. Not sure whether writing an empty string would work, but we go with actually writing some data.
For what it's worth, the default value of this dummy is a space (` `) in this lower-level API, but in `build_token_cache(...)` we (have to) explicitly define it as a valid JSON payload `{}`.

msal_extensions/token_cache.py
rayluo: @chlowell was eloquent during our offline discussion:
> I feel the helper doesn't really help me because it doesn't hide the particulars of persistence. To use it, I must learn which options apply on which platforms, and how; which ones are truly optional. Then I either pass all options at once or call the helper conditionally by platform. I prefer to write the conditional code because it would be easier for future generations to understand. At that point I may as well just import and configure persistence classes directly, that's nicely explicit.

The two latest commits demoted the `build_persistence()` and `build_token_cache()` into their own sample, i.e. removed them from API surface.

@chlowell please review again. If you like it, please approve. :-)

msal_extensions/libsecret.py
```diff
@@ -0,0 +1,79 @@
...
+    except (gi.repository.GLib.Error, AssertionError):
+        message = (
+            "libsecret did not perform properly. Please refer to "
+            "https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/Encryption-on-Linux")  # pylint: disable=line-too-long
+        logger.exception(message)  # This log contains trace stack for debugging
+        logger.warning(message)  # This is visible by default
+        raise
+
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
rayluo: Hmm, it was the very first test case I used when TDDing a proof-of-concept. And then it was left here as a test case for this literally self-contained libsecret.py module (which could potentially be useful in itself).
I have no real objection to remove it, but my question is, is there any practical (rather than aesthetical) reason that we should remove it?

msal_extensions/persistence.py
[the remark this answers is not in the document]
rayluo: All the dummies are removed in latest iteration.

msal_extensions/persistence.py
```diff
@@ -0,0 +1,225 @@
...
+        """Initialization could fail due to unsatisfied dependency"""
+        if not (service_name and account_name):  # It would hang on OSX
+            raise ValueError("service_name and account_name are required")
+        from .osx import Keychain  # pylint: disable=import-outside-toplevel
+        self._file_persistence = FilePersistence(location)  # Favor composition
```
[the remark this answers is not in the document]
rayluo: I can understand your point. The `location` was named this way, and even at the first position, only because historically the entire persistence family were subclasses of `FilePersistence`. Given that it is no longer the case, sure it is good idea to at least rename `location` to something like `flag_path`? Naming is hard.

But then your other message saying your attempting `location=None` usage is an even bigger topic. Normally a required parameter is not expected to accept `None` unless it was documented otherwise. But did you hint that you foresee a usage to skip that flag file all together? I'm fine to choose to allow `flag_path=None` in this API surface (or to allow it implicitly inside implementation), but it would also mean the burden is on your side (i.e. Azure SDK) to remember to provide a flag file in order for cross-process timestamp to work. Is that what you want?

msal_extensions/persistence.py
```diff
@@ -0,0 +1,225 @@
...
+    is_encrypted = True
+
+    def __init__(self, location, schema_name, attributes, dummy=" ", **kwargs):
+        """Initialization could fail due to unsatisfied dependency"""
+        # pylint: disable=import-outside-toplevel
+        from .libsecret import (  # This uncertain import is deferred till runtime
+            LibSecretAgent, verify)
+        verify()
+        if not attributes:
+            raise ValueError("attributes is required")
```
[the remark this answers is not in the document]
rayluo: Good catch. Will add more doc in docstring and then improve this placeholder.

msal_extensions/persistence.py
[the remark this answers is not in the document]
chlowell: What I meant regarding `location=None` (really, `not location`) is something simpler: you raise `ValueError` when `service_name` or `account_name` is falsey, to prevent a hang. Maybe it's worth doing the same for `location`, to get a better error message?

msal_extensions/libsecret.py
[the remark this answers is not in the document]
chlowell: No, I point it out only because it doesn't seem useful for end users, and I habitually scrutinize my own PRs to avoid merging my debugging entrypoints 😊

msal_extensions/persistence.py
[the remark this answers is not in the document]
rayluo: Renamed it to `signal_location`. Also, a `signal_location = None` input would trigger a ValueError exception this time.

PS: The term "signal" did not come up from nowhere. During some of early days brainstorming, yet another file named "your_token_cache.bin.signal" was briefly considered, alongside with the developer-provided "your_token_cache.bin" and the auto-generated "your_token_cache.bin.lock". Although we ended up not explicitly maintaining such 3-files setup, the "your_token_cache.bin" provided by app developer will indeed be used as the signal file, when running on OSX and Linux Desktop.

msal_extensions/persistence.py
```diff
@@ -0,0 +1,225 @@
...
+
+class KeychainPersistence(BasePersistence):
+    """A generic persistence with data stored in,
+    and protected by native Keychain libraries on OSX"""
+    is_encrypted = True
+
+    def __init__(self, location, service_name, account_name, dummy=" "):
+        """Initialization could fail due to unsatisfied dependency"""
+        if not (service_name and account_name):  # It would hang on OSX
+            raise ValueError("service_name and account_name are required")
```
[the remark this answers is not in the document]
rayluo: Added a ValueError for this.

PR #47, opened 2020-04-17
Only require PyGObject at runtime

rayluo:
Per request from Azure SDK team, we no longer require `PyGObject` package at installation time. For this change, we also fine tune the run-time errors.

@chlowell please see if this is what you want.

### DIFF
--- a/msal_extensions/libsecret.py
+++ b/msal_extensions/libsecret.py
@@ -15,15 +15,31 @@ Alternatively, you could skip Cairo & PyCairo, but you still need to do all thes
 """
 import logging
 
-import gi  # https://pygobject.readthedocs.io/en/latest/getting_started.html
-
-# pylint: disable=no-name-in-module
-gi.require_version("Secret", "1")  # Would require a package gir1.2-secret-1
-# pylint: disable=wrong-import-position
-from gi.repository import Secret  # Would require a package gir1.2-secret-1
-
-
 logger = logging.getLogger(__name__)
+
+try:
+    import gi  # https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/Encryption-on-Linux
+except ImportError:
+    logger.exception(
+        """Runtime dependency of PyGObject is missing.
+Depends on your Linux distro, you could install it system-wide by something like:
+    sudo apt install python3-gi python3-gi-cairo gir1.2-secret-1
+If necessary, please refer to PyGObject's doc:
+https://pygobject.readthedocs.io/en/latest/getting_started.html
+""")
+    raise
+
+try:
+    # pylint: disable=no-name-in-module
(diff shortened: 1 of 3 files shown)


msal_extensions/libsecret.py
```diff
@@ -111,9 +127,12 @@ def trial_run():
...
         agent.clear()
     except (gi.repository.GLib.Error, AssertionError):
-        message = (
-            "libsecret did not perform properly. Please refer to "
-            "https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/Encryption-on-Linux")  # pylint: disable=line-too-long
+        message = """libsecret did not perform properly.
+* If you encountered error "Remote error from secret service:
+  org.freedesktop.DBus.Error.ServiceUnknown",
+  you may need to install gnome-keyring package.
+* Headless mode (such as in an ssh session) is not supported.
```
chlowell, line 134: Just for my own understanding: is this because GNOME tries to display UI?

msal_extensions/libsecret.py
```diff
@@ -15,16 +15,32 @@
...
-# pylint: disable=no-name-in-module
-gi.require_version("Secret", "1")  # Would require a package gir1.2-secret-1
-# pylint: disable=wrong-import-position
-from gi.repository import Secret  # Would require a package gir1.2-secret-1
-
-
 logger = logging.getLogger(__name__)
 
+try:
+    import gi  # https://pygobject.readthedocs.io/en/latest/getting_started.html
```
chlowell, line 21: If someone doesn't have logging configured, they won't see the explanation logged below, but this line will be part of the traceback. It could be better to link to an explanation in this library's documentation.

rayluo: Glad that you've also noticed that this file was deliberately arranged to have inline comment helpful in error trace.

I've also adopted your suggestion to link to a wiki page in this repo (which can be further refined at a later time).

msal_extensions/libsecret.py
```diff
@@ -15,16 +15,32 @@
...
+except ImportError:
+    logger.exception(
```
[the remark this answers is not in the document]
rayluo: That would not be possible this time, since we've demoted the `build_persistence(..., fallback_to_plaintext=False)` helper to become a sample only. :-)

msal_extensions/libsecret.py
rayluo: In fact, the UI display seems only needed/visible at the first interaction during a desktop session, after that it works silently. My understanding is the secret service is somehow only activated in an X session, otherwise you will see this error:
> gi.repository.GLib.GError: g-io-error-quark: Cannot autolaunch D-Bus without X11 $DISPLAY (0)

PR #49, opened 2020-04-22
Use more precise dependency declaration

rayluo:
Per [a discussion](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/43#issuecomment-617739430) and its subsequent off-line discussion with @bluca , we realize that package maintainers on Linux might not want to aggressively upgrade a dependency to its latest version. So, in this PR, we make it clear that the portalocker 1.6.0 dependency requirement is only needed on Windows.


bluca:
Thank you!

PR #50, opened 2020-04-23
Restore compatibility with portalocker < 1.4.0

bluca:
Support for passing parameters through to the open syscall was added in
version 1.4.0. Check before using this feature to keep backward compatibility.

https://github.com/WoLpH/portalocker/commit/721b3309c2

The dependency would cut off the current Debian Stable and Ubuntu before 19.10

### DIFF
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -3,6 +3,7 @@ import os
 import sys
 import errno
 import portalocker
+from distutils.version import LooseVersion
 
 
 class CrossPlatLock(object):
@@ -12,6 +13,8 @@ class CrossPlatLock(object):
     """
     def __init__(self, lockfile_path):
         self._lockpath = lockfile_path
+        # Support for passing through arguments to the open syscall was added in v1.4.0
+        open_kwargs = {'buffering': 0} if LooseVersion(portalocker.__version__) >= LooseVersion("1.4.0") else {}
         self._lock = portalocker.Lock(
             lockfile_path,
             mode='wb+',
@@ -20,7 +23,7 @@ class CrossPlatLock(object):
             # More information here:
             # https://docs.python.org/3/library/fcntl.html#fcntl.lockf
             flags=portalocker.LOCK_EX | portalocker.LOCK_NB,
-            buffering=0)
+            **open_kwargs)
 
     def __enter__(self):
         file_handle = self._lock.__enter__()


bluca:
Do you intend to keep support for Python 2 even though it's EOL? If so let me know and I'll rework this

msal_extensions/cache_lock.py
```diff
@@ -12,6 +13,8 @@ class CrossPlatLock(object):
     """
     def __init__(self, lockfile_path):
         self._lockpath = lockfile_path
+        # Support for passing through arguments to the open syscall was added in v1.4.0
+        open_kwargs = {'buffering': 0} if inspect.signature(portalocker.Lock).parameters.get('file_open_kwargs') else {}
```
rayluo, line 17: You could just use `portalocker.__version__` for the version check. Not bullet-proof, but good enough. And it won't break our Python 2 support.

msal_extensions/cache_lock.py
```diff
@@ -12,6 +13,8 @@ class CrossPlatLock(object):
     """
     def __init__(self, lockfile_path):
         self._lockpath = lockfile_path
+        # Support for passing through arguments to the open syscall was added in v1.4.0
```
rayluo, line 16: Nice catch. In hindsight, we should have declare the dependency of `portalocker>=1.4.0` since day 1.

Are you saying "portalocker>=1.4.0 would cut off the current Debian Stable and Ubuntu before 19.10"? Why so?

If that is really the case, I'm fine with your proposed run-time detection, for now. (But if in the future we would hear bug reports saying [the default `buffering=-1` behavior](https://docs.python.org/3/library/functions.html#open) causes trouble, we would then switch to have a hard dependency on portalocker>=1.4.0

rayluo:
@bluca If at all possible, please prioritize this fix. We found and fixed a separated bug in 0.2.0+ series, and we would like to have a new bugfix release 0.2.2 first thing next week. Ideally your improvement here can also go in in that release.

msal_extensions/cache_lock.py
bluca: Ok, done - I was trying to be clever, but there's really no need

msal_extensions/cache_lock.py
bluca: https://packages.ubuntu.com/search?keywords=python3-portalocker&searchon=names&suite=all&section=all
https://packages.debian.org/search?keywords=python3-portalocker&searchon=names&suite=all&section=all

msal_extensions/cache_lock.py
```diff
@@ -3,6 +3,7 @@
 import sys
 import errno
 import portalocker
+from distutils.version import LooseVersion
```
rayluo, line 6: Interesting. I did not know this. The `distutils.version` was [barely mentioned but not documented in Python's doc](https://docs.python.org/3/distutils/apiref.html#module-distutils.version). With that specific name, I managed to find  [doc for `LooseVersion`](http://epydoc.sourceforge.net/stdlib/distutils.version.LooseVersion-class.html), and accidentally saw [this issue report](https://github.com/postmarketOS/pmbootstrap/issues/342).

Anyway, we can go with this, for now.

#LearnSomethingNewEveryday

bluca: The alternative is setuptools, which is less hacky but adds a new runtime dependency

rayluo: Agree with your choice to avoid yet another dependency.

PR #52, opened 2020-04-24
Writing of protected cache in bytes for Windows 10 DPAPI 

abhidnya13:


msal_extensions/persistence.py
```diff
@@ -114,12 +114,13 @@ def __init__(self, location, entropy=''):
...
-            self._dp_agent.protect(content))
+        with open(self._location, 'wb+') as handle:
+            handle.write(self._dp_agent.protect(content))
 
     def load(self):
-        return self._dp_agent.unprotect(
-            super(FilePersistenceWithDataProtection, self).load())
+        with open(self._location, 'rb') as handle:
+            return self._dp_agent.unprotect(
+                handle.read())
```
rayluo, line 123:
```suggestion
        # type: () -> str
        with open(self._location, 'rb') as handle:
            return self._dp_agent.unprotect(handle.read())
```

Given that we are getting into the nasty difference of string and bytes anyway, let's also include a type hint here. Same suggestion for `save()` too. In the near future (but not in this PR), we will pick up another task for documentation, and describe this in details there.

Also, since this is not automated, please manually test [this test case](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/0.2.1/tests/test_persistence.py#L35) and [that sample](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/0.2.1/sample/persistence_sample.py), on both Python 2.7 and one version of Python 3, we need to see all pass in such 2x2 matrix. Please ack.

abhidnya13: Manually tested the test case and the sample on Python 2.7 and 3.7 and it works fine

PR #53, opened 2020-04-24
Add Readme usage

navyasric:


README.md
```diff
@@ -1,5 +1,87 @@
...
+
+This is available for Windows, Mac and Linux. 
+- Windows - [DPAPI](https://docs.microsoft.com/en-us/dotnet/standard/security/how-to-use-data-protection) is used for encryption.
+- MAC - The MAC KeyChain is used and encryption is built in.
+- Linux - [LibSecret](https://wiki.gnome.org/Projects/Libsecret) is used for encryption. Any place where libsecret is not available encryption is not supported.
+
+> Note: This library provides cache persistance support for Public client applications such as Desktop apps only. Web applications are recommended to persist the cache in memory or in Redis. Take a look at this [webapp sample](https://github.com/Azure-Samples/ms-identity-python-webapp).
```
rayluo, line 13:
```suggestion
> Note: This library provides cache persistance support for Public client applications such as Desktop apps only. Web applications are recommended to persist the cache in session. Take a look at this [webapp sample](https://github.com/Azure-Samples/ms-identity-python-webapp).

```

I think `session` is a more generic yet more precise term here. Our sample is not using cache in memory, and is not demonstrating Redis.

README.md
```diff
@@ -1,5 +1,87 @@
...
+
+## Usage
+
+The Microsoft Authentication Extensions library provides the `PersistedTokenCache` which accepts a platform-dependent persistence instance. This token cache can then be used to instantiate the `PublicClientApplication` in MSAL Python. 
+
+The token cache includes a file lock, and auto-reload behavior under the hood.
+
+
+
+Here is an example of this pattern for multiple platforms:
```
rayluo, line 37: Perhaps also provide a link to the real sample, in case that sample would receive some new enhancement or bugfix in future, while we might not remember to update the code snippet inside this README?

UPDATED: Oh, after finishing reading this entire section, I realize that you did provide a link to that sample at the end. Thanks! But would it still be better to provide that link up front?

```suggestion
Here is an example of this pattern for multiple platforms
(and its latest implementation is available
[here](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/dev/sample/persistence_sample.py)
):

```

PR #58, opened 2020-05-11
Performance Testing

abhidnya13:
Reference:

* [MSAL EX .Net's Test lock under stress](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/82/files)
* [MSAL EX Java's refactoring](https://github.com/AzureAD/microsoft-authentication-extensions-for-java/commit/7d789a11f005dca07ad7b8c20060664f35ab4851)

@bgavrilMS mentioned:

> produce a simple console app that
> * accepts 3 inputs: string protectedFilePath, string lockFilePath and int millisecondsToSleep
> * does the following:
> ```
>	acquire cross-plat lock 
>	write " > process_id" to protectedFile
>	sleep 
>	write " < process_id" to protected File
> ```

### DIFF
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -34,6 +34,6 @@ class CrossPlatLock(object):
             # likely that another process has raced this one and ended up clearing or locking the
             # file for itself.
             os.remove(self._lockpath)
-        except OSError as ex:
+        except OSError as ex:  # pylint: disable=invalid-name
             if ex.errno != errno.ENOENT and ex.errno != errno.EACCES:
                 raise
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -2,6 +2,7 @@
 import os
 import sys
 import errno
+import time
 import portalocker
 
 
@@ -10,8 +11,9 @@ class CrossPlatLock(object):
     resource. This is specifically written to interact with a class of the same name in the .NET
     extensions library.
     """
-    def __init__(self, lockfile_path):
+    def __init__(self, lockfile_path, time_interval_file_path):
         self._lockpath = lockfile_path
+        self.time_interval_file_path = time_interval_file_path
         self._lock = portalocker.Lock(
             lockfile_path,
             mode='wb+',
@@ -20,15 +22,19 @@ class CrossPlatLock(object):
             # More information here:
             # https://docs.python.org/3/library/fcntl.html#fcntl.lockf
             flags=portalocker.LOCK_EX | portalocker.LOCK_NB,
+            timeout=90,
             buffering=0)
 
     def __enter__(self):
+        self.start_time = str(time.time())
         file_handle = self._lock.__enter__()
         file_handle.write('{} {}'.format(os.getpid(), sys.argv[0]).encode('utf-8'))
         return file_handle
 
     def __exit__(self, *args):
         self._lock.__exit__(*args)
+        with open(self.time_interval_file_path, 'a+') as handle:
+            handle.write(self.start_time + "-" + str(time.time())+ "\n")
         try:
             # Attempt to delete the lockfile. In either of the failure cases enumerated below, it is
             # likely that another process has raced this one and ended up clearing or locking the
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -11,7 +11,7 @@ class CrossPlatLock(object):
     resource. This is specifically written to interact with a class of the same name in the .NET
     extensions library.
     """
-    def __init__(self, lockfile_path, timeout=5):
+    def __init__(self, lockfile_path):
         self._lockpath = lockfile_path
         # Support for passing through arguments to the open syscall was added in v1.4.0
         open_kwargs = {'buffering': 0} if LooseVersion(portalocker.__version__) >= LooseVersion("1.4.0") else {}
@@ -23,7 +23,6 @@ class CrossPlatLock(object):
             # More information here:
             # https://docs.python.org/3/library/fcntl.html#fcntl.lockf
             flags=portalocker.LOCK_EX | portalocker.LOCK_NB,
-            timeout=timeout,
             **open_kwargs)
 
     def __enter__(self):
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -2,8 +2,8 @@
 import os
 import sys
 import errno
-import time
 import portalocker
+from distutils.version import LooseVersion
 
 
 class CrossPlatLock(object):
@@ -11,9 +11,10 @@ class CrossPlatLock(object):
     resource. This is specifically written to interact with a class of the same name in the .NET
     extensions library.
     """
-    def __init__(self, lockfile_path, time_interval_file_path):
+    def __init__(self, lockfile_path, timeout=5):
         self._lockpath = lockfile_path
-        self.time_interval_file_path = time_interval_file_path
+        # Support for passing through arguments to the open syscall was added in v1.4.0
+        open_kwargs = {'buffering': 0} if LooseVersion(portalocker.__version__) >= LooseVersion("1.4.0") else {}
         self._lock = portalocker.Lock(
             lockfile_path,
             mode='wb+',
@@ -22,19 +23,16 @@ class CrossPlatLock(object):
             # More information here:
             # https://docs.python.org/3/library/fcntl.html#fcntl.lockf
             flags=portalocker.LOCK_EX | portalocker.LOCK_NB,
-            timeout=90,
-            buffering=0)
+            timeout=timeout,
+            **open_kwargs)
 
     def __enter__(self):
-        self.start_time = str(time.time())
         file_handle = self._lock.__enter__()
         file_handle.write('{} {}'.format(os.getpid(), sys.argv[0]).encode('utf-8'))
         return file_handle
 
     def __exit__(self, *args):
         self._lock.__exit__(*args)
-        with open(self.time_interval_file_path, 'a+') as handle:
-            handle.write(self.start_time + "-" + str(time.time())+ "\n")
         try:
             # Attempt to delete the lockfile. In either of the failure cases enumerated below, it is
             # likely that another process has raced this one and ended up clearing or locking the
--- a/tests/cache_file_generator.py
+++ b/tests/cache_file_generator.py
@@ -1,3 +1,11 @@
+"""
+This is a console application which is to be used for cross-platform lock performance testing.
+The app will acquire lock for the cache file, log the process id and free the lock
+It takes in two arguments - cache file path and the sleep interval.
+The cache file path is the path of cache file.
+The sleep interval is the time in seconds for which the lock is held by a process.
+"""
+
 import logging
 import os
 import sys
--- a/tests/cache_file_generator.py
+++ b/tests/cache_file_generator.py
@@ -1,6 +1,9 @@
 """
+Usage: cache_file_generator.py cache_file_path sleep_interval
+
 This is a console application which is to be used for cross-platform lock performance testing.
-The app will acquire lock for the cache file, log the process id and free the lock
+The app will acquire lock for the cache file, log the process id and then release the lock.
+
 It takes in two arguments - cache file path and the sleep interval.
 The cache file path is the path of cache file.
 The sleep interval is the time in seconds for which the lock is held by a process.
@@ -33,5 +36,8 @@ def _acquire_lock_and_write_to_cache(cache_location, sleep_interval):
 
 
 if __name__ == "__main__":
+    if len(sys.argv) < 3:
+        print(__doc__)
+        sys.exit(0)
     _acquire_lock_and_write_to_cache(sys.argv[1], float(sys.argv[2]))
 
--- a/tests/cache_file_generator.py
+++ b/tests/cache_file_generator.py
@@ -1,0 +1,29 @@
+import logging
+import os
+import sys
+import time
+
+from portalocker import exceptions
+
+from msal_extensions import FilePersistence, CrossPlatLock
+
+
+def _acquire_lock_and_write_to_cache(cache_location, sleep_interval):
+    cache_accessor = FilePersistence(cache_location)
+    lock_file_path = cache_accessor.get_location() + ".lockfile"
+    try:
+        with CrossPlatLock(lock_file_path):
+            data = cache_accessor.load()
+            if data is None:
+                data = ""
+            data += "< " + str(os.getpid()) + "\n"
+            time.sleep(sleep_interval)
+            data += "> " + str(os.getpid()) + "\n"
+            cache_accessor.save(data)
+    except exceptions.LockException as e:
+        logging.warning("Unable to acquire lock %s", e)
+
+
+if __name__ == "__main__":
+    _acquire_lock_and_write_to_cache(sys.argv[1], float(sys.argv[2]))
+
--- a/tests/test_cache_lock_file_perf.py
+++ b/tests/test_cache_lock_file_perf.py
@@ -31,27 +31,20 @@ def _validate_result_in_cache(expected_entry_count, cache_location):
             else:
                 assert tag == '<', "Opening bracket not found"
                 prev_process_id = process_id
-
-    assert count <= expected_entry_count * 2, "File content corrupted"
-    if count < expected_entry_count * 2:
-        logging.warning("Starvation detected")
+    return count
 
 
 def _acquire_lock_and_write_to_cache(cache_location, sleep_interval=0):
     cache_accessor = FilePersistence(cache_location)
     lock_file_path = cache_accessor.get_location() + ".lockfile"
-    try:
-        with CrossPlatLock(lock_file_path):
-            data = cache_accessor.load()
-            if data is None:
-                data = ""
-            data += "< " + str(os.getpid()) + "\n"
-            time.sleep(sleep_interval)
-            data += "> " + str(os.getpid()) + "\n"
-            cache_accessor.save(data)
-    except Exception as e:
-        logging.warning("Exception raised %s", e)
-        sys.exit(3)
+    with CrossPlatLock(lock_file_path):
+        data = cache_accessor.load()
+        if data is None:
+            data = ""
+        data += "< " + str(os.getpid()) + "\n"
+        time.sleep(sleep_interval)
+        data += "> " + str(os.getpid()) + "\n"
+        cache_accessor.save(data)
 
 
 def _run_multiple_processes(no_of_processes, cache_location, sleep_interval):
@@ -68,23 +61,23 @@ def _run_multiple_processes(no_of_processes, cache_location, sleep_interval):
 
     for process in processes:
         process.join()
-        if process.exitcode == 3:
-            raise Exception
 
 
 def test_multiple_processes_without_timeout_exception(cache_location):
     num_of_processes = 20
     sleep_interval = 0
     _run_multiple_processes(num_of_processes, cache_location, sleep_interval)
-    _validate_result_in_cache(num_of_processes, cache_location)
+    count = _validate_result_in_cache(num_of_processes, cache_location)
+    assert count == num_of_processes * 2
     os.remove(cache_location)
 
 
 def test_multiple_processes_with_timeout_exception_raised(cache_location):
     num_of_processes = 10
     sleep_interval = 1
-    with pytest.raises(Exception):
-        _run_multiple_processes(
-            num_of_processes, cache_location, sleep_interval)
+    _run_multiple_processes(
+        num_of_processes, cache_location, sleep_interval)
+    count = _validate_result_in_cache(num_of_processes, cache_location)
+    assert count < num_of_processes * 2, "Should observe starvation"
     os.remove(cache_location)
 
--- a/tests/test_cache_lock_file_perf.py
+++ b/tests/test_cache_lock_file_perf.py
@@ -1,14 +1,11 @@
-import logging
 import multiprocessing
 import os
 import shutil
 import tempfile
-import time
 
-from portalocker import exceptions
 import pytest
 
-from msal_extensions import FilePersistence, CrossPlatLock
+from cache_file_generator import _acquire_lock_and_write_to_cache
 
 
 @pytest.fixture
@@ -35,22 +32,6 @@ def _validate_result_in_cache(cache_location):
                 assert tag == '<', "Opening bracket not found"
                 prev_process_id = process_id
     return count
-
-
-def _acquire_lock_and_write_to_cache(cache_location, sleep_interval):
-    cache_accessor = FilePersistence(cache_location)
-    lock_file_path = cache_accessor.get_location() + ".lockfile"
-    try:
-        with CrossPlatLock(lock_file_path):
-            data = cache_accessor.load()
-            if data is None:
-                data = ""
-            data += "< " + str(os.getpid()) + "\n"
-            time.sleep(sleep_interval)
-            data += "> " + str(os.getpid()) + "\n"
-            cache_accessor.save(data)
-    except exceptions.LockException as e:
-        logging.warning("Unable to acquire lock %s", e)
 
 
 def _run_multiple_processes(no_of_processes, cache_location, sleep_interval):
--- a/tests/test_cache_lock_file_perf.py
+++ b/tests/test_cache_lock_file_perf.py
@@ -72,7 +72,7 @@ def _run_multiple_processes(no_of_processes, cache_location, sleep_interval):
 
 
 def test_multiple_processes_without_timeout_exception(cache_location):
-    num_of_processes = 80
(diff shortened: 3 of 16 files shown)


msal_extensions/cache_lock.py
```diff
@@ -20,15 +22,19 @@ def __init__(self, lockfile_path):
...
     def __enter__(self):
+        self.start_time = str(time.time())
         file_handle = self._lock.__enter__()
         file_handle.write('{} {}'.format(os.getpid(), sys.argv[0]).encode('utf-8'))
         return file_handle
 
     def __exit__(self, *args):
         self._lock.__exit__(*args)
+        with open(self.time_interval_file_path, 'a+') as handle:
+            handle.write(self.start_time + "-" + str(time.time())+ "\n")
```
rayluo, line 37: [MSAL EX Java implemented it as part of the test suite](https://github.com/AzureAD/microsoft-authentication-extensions-for-java/commit/7d789a11f005dca07ad7b8c20060664f35ab4851#diff-62402c24972c337298ee12d0616b6aedR44). We should not modify our real `cache_lock.py` code path for this test.

[MSAL EX .Net did not implement this part](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/82/files#diff-501fa80648fb3dcda4b2ce004dd24bebR45) at all. I think that is a reasonable move. Our offline conversation confirmed that in this PoC, the test failure could show up in either `validate_thread_locking_file()` or `validate_result_in_cache()`, so they seem the same problem, we do not have to test 2 files for the same underlying issue.

I would suggest to roll back the changes to this file, in order to have this PR eventually merged in.

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,110 @@
...
+        if prev:
+            if start < prev:
+                print("overlapping")
+        prev = start
+    print("success")
+    return
+
+
+def validate_result_in_cache(num):
+    with open("C:/Users/abpati/Desktop/test/msal.cache") as handle:
```
rayluo, line 33: Next step is to not hard code any specific path, so that this test case would be able to be run at anywhere, including CI/CD.

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,110 @@
...
+    for i in range(num_of_processes):
+        t = multiprocessing.Process(target=do_something, args=(i,))
+        processes.append(t)
+
+    for i in processes:
+        i.start()
+
+    for i in processes:
+        i.join()
+    validate_thread_locking_file(num_of_processes)
```
rayluo, line 105: Please refactor the test driver to be a stand alone CLI script, and then have the test validation as a unit test case.

abhidnya13:
Yes, this definitely is a very rough draft and I would be making a lot more changes, stay tuned !

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,110 @@
...
+
+def validate_thread_locking_file(num):
+    with open("C:/Users/abpati/Desktop/test/lockfilelog.txt") as handle:
+        data = handle.read()
+    time_list = []
+    for line in data.split("\n"):
+        start, end = line.split("-")
+        time_list.append([start, end])
+    if len(time_list) != num:
+        print("Threads count dont match ")
```
bgavrilMS, line 15: nit: typo...

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,110 @@
+import multiprocessing
+import threading
+import time
+from msal_extensions import FilePersistence, CrossPlatLock
+
+
+def validate_thread_locking_file(num):
+    with open("C:/Users/abpati/Desktop/test/lockfilelog.txt") as handle:
```
bgavrilMS, line 8: won't work on Linux :) Consider putting this in the same directory as the py script or in some other directory that is known to exist

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,110 @@
+import multiprocessing
+import threading
+import time
+from msal_extensions import FilePersistence, CrossPlatLock
+
+
+def validate_thread_locking_file(num):
```
bgavrilMS, line 7: `num` is a pretty common noun in programming. Consider using longer, more descriptive names for variables. `expectedThreadCount` in this case would make this method more readable.

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,110 @@
...
+    with open("C:/Users/abpati/Desktop/test/lockfilelog.txt") as handle:
+        data = handle.read()
+    time_list = []
+    for line in data.split("\n"):
+        start, end = line.split("-")
+        time_list.append([start, end])
+    if len(time_list) != num:
+        print("Threads count dont match ")
+        return
+    time_list.sort()
```
bgavrilMS, line 17: Not sure I understand why this list needs to be sorted?

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,110 @@
...
+    if ">" != prev_tag:
+        print("Failed")
+        return
+    if count != num:
+        print("Failed")
+        return
+    print("sucess")
+
+
+def do_something(tid):
```
bgavrilMS, line 62: `do_something` makes the code hard to read, as it is too general. Perhaps `write_to_protected_file` ?

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,110 @@
...
+    lock_intervals_file_path = "C:/Users/abpati/Desktop/test/lockfilelog.txt"
+    lock_file_path = "C:/Users/abpati/Desktop/test/msal.cache.lockfile"
+    cache_accessor = FilePersistence("C:/Users/abpati/Desktop/test/msal.cache")
+    with CrossPlatLock(lock_file_path, lock_intervals_file_path):
+        thr = str(tid)
+        data = cache_accessor.load()
+        if data is None:
+            data = ""
+        data += "< " + thr + "\n"
+        time.sleep(0.1)
```
bgavrilMS, line 72: Did you notice different behvior with 1s instead of 0.1s here ? I think Java and .NET used 1s to make sure we get higher concurrency..

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,110 @@
...
+        data += "> " + thr + "\n"
+        cache_accessor.save(data)
+
+
+def multiple_threads():
```
bgavrilMS, line 77: @rayluo - it's interesting that both Java and Python ensure both cross-process and cross-thread concurrency. In .NET, we only ensure cross-process, because MSAL takes care of cross-thread. Just an observation that you may be locking too much...

tests/test_cache_lock_file_perf.py
rayluo: This parameter should become a parameter of this method, so that we can better customize the configuration in test automation.

abhidnya13:
Thanks for reviewing @bgavrilMS , I am supposed to make changes, created a draft PR as Ray mentioned it would be good to have a draft PR out for now and then if possible he can test it on his Linux machine as well. Will push updates soon on a more polished implementation. However I did test it on my Mac and I noticed that processes ~50 is when it starts to give cache errors.  I also noticed on Mac that it fails after ~400 threads but as you pointed out MSAL ensures thread safety would possibly mean we would not get these errors.

tests/test_cache_lock_file_perf.py
rayluo: I also had to confirm with @SomkaPe during our daily stand-up yesterday to understand why Java implementation did this. His answer was basically "it simple to also implement multi-thread testing, so we just did it".

My take on this:

> "you may be locking too much..."

Originally I thought @bgavrilMS you mean our production implementation might lock too much. That would not be the case, because our real code path handles lock in exact same way, regardless of being used in multi-thread setup or multi-process setup.

And then I realize you actually mean "we may be *applying* locking too much *during our test*". It is true that, from MSAL EX's perspective, there would be no real need to test multi-thread scenario, because this MSAL EX is meant to be used in multi-process scenario. To that end, perhaps this PR would eventually only automate a multi-process test case with mild workload, which serves as a "known safe limit".

But then, given that part of the motivation of this test is to prove whether that hard limit exists, it doesn't harm to have a multi-thread capability (which is more efficient than multi-process), for us to manually run it and quickly hit that upper bound.

msal_extensions/cache_lock.py
```diff
@@ -11,7 +11,7 @@ class CrossPlatLock(object):
     resource. This is specifically written to interact with a class of the same name in the .NET
     extensions library.
     """
-    def __init__(self, lockfile_path):
+    def __init__(self, lockfile_path, timeout=5):
```
rayluo, line 14: But why do we need to customize such timeout, even in our stress test? Aren't we wanna see when it would fail under NORMAL setup?

abhidnya13: So default timeout behaviour will basically make even 5 processes trying to access cache at the same time fail as portalocker will throw an exception. If this is configurable its useful if the use case for someone expects a load of more processes than 5 . I would say we should expose it to be set in the API but if we decide we dont want to expose it, we can guarantee only <= 5 processes to access cache roughly around the same time. Thats the limitation of the library.

abhidnya13: So basically this stress test would not even fail because Mac and Linux locks are not reliable after n processes. This would fail because there was a timeout.

rayluo: Thanks for this new info. That is exactly what I have been wondering, "why was there a cache failure in the first place? Shouldn't any lock mechanism raise exception when a lock isn't available, instead of silently bypassing the lock?". In that sense, that unable-to-acquire-lock exception is an expected feature, rather than a failure.

Before we make any adjustment (ideally we shouldn't have to), we want to know the behavior of our MSAL EX under default configuration. And then we go from there.

Whether we expose that timeout as a parameter, is a different topic. I doubt it would be meaningful in practice. The default 5 second timeout already makes sense.

abhidnya13: After a quick run on my windows machine, as the sleep interval is set as 1 sec , 5 processes is what the maximum processes can access, the lock exception is raised with 6 processes. But then again this number will depend on the sleep interval. 

I am not sure if 5 secs makes sense. It definitely does for some cases like the test case here for 5 processes, but what if there is a use case of more than 5 processes our library would fail and they would need some extra application logic to handle the timeout. In those cases if this is an expected behavior for the application instead of adding that handling logic we can have a configurable timeout instead. We would still have a default timeout of 5 secs if they don't want to configure it.

May be if we have an idea about the normal expected workload, this can be set to default to some value like 5 secs. And then we can do quick tests around how many processes are able to work under this default configuration depending on how long the lock is held.

I believe it is a configurable setting for MSAL Java and MSAL .net @bgavrilMS can confirm. Last I checked with them, java has a 6 sec default time out behavior and .net has default timeout of 60 secs but it is configurable.

rayluo: Based on the off-line discussion among us and @SomkaPe just now, let's try using the default 5 seconds timeout here (*), and:

* Have one automated test case to use `sleep_interval` as 1 second, and assert we will encounter some unable-to-acquire-lock exception when under 5+ concurrent processes. This situation is considered as a test PASS, and it would mean the locking mechanism works as expected, at least when under low load.

* Have another automated test case which lower the `sleep_interval` to very close to, or even equal to, 0 second, and to observe everything works under some perhaps ~20 processes? This test pass would mean our MSAL EX would be safe under such artificial load.

(*) If the above 2 tests pass, we do not have to expose this `timeout` parameter here.

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,80 @@
...
+            tag, process_id = line.split(" ")
+            if prev_process_id is not None:
+                assert process_id == prev_process_id, "Process overlap found"
+                assert tag == '>', "Process overlap_found"
+                prev_process_id = None
+            else:
+                assert tag == '<', "Opening bracket not found"
+                prev_process_id = process_id
+
+    assert count == expected_entry_count*2, "No of processes don't match"
```
rayluo, line 33: Add an `import logging` at the beginning, and then change this line into:
```suggestion
    assert count <= expected_entry_count*2, "File content corrupted"
    if count < expected_entry_count*2:
        logging.warning("Starvation detected")
```

After that, rerun the test 3~5 times to see if it still "fails". If not, then you can tentatively try ramping up the `num_of_processses` from its current value `10` to `20`, rinse and repeat. And then, perhaps 30, 50, 70..., if it all goes well, you can eventually go all the way up to 100, which was the value used in .Net test. (What number was used by Java, again?)

And then we discuss.

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,80 @@
...
+def _run_multiple_processes(no_of_processes, cache_location, sleep_interval):
+    open(cache_location, "w+")
+    processes = []
+    for i in range(no_of_processes):
+        t = multiprocessing.Process(
+            target=_acquire_lock_and_write_to_cache,
+            args=(cache_location, sleep_interval))
+        processes.append(t)
+
+    for i in processes:
```
rayluo, line 58: Nit:
```suggestion
    for process in processes:
```

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,83 @@
...
+    _run_multiple_processes(num_of_processes, cache_location, sleep_interval)
+    _validate_result_in_cache(num_of_processes, cache_location)
+    os.remove(cache_location)
+
+
+def test_multiple_processes_with_timeout_exception_raised(cache_location):
+    num_of_processes = 6
+    sleep_interval = 1
+    with pytest.raises(Exception):
+        assert _run_multiple_processes(
```
rayluo, line 80:
```suggestion
        _run_multiple_processes(
```

rayluo: A follow-up: @abhidnya13 , interestingly, this supposed-to-be-irrelevant minor change causes a test failure now, which exposes a much bigger issue we overlooked. This test case was not right. Please adjust the `with pytest.raises(Exception)`.

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,90 @@
...
+    sleep_interval = 0
+    _run_multiple_processes(num_of_processes, cache_location, sleep_interval)
+    _validate_result_in_cache(num_of_processes, cache_location)
+    os.remove(cache_location)
+
+
+def test_multiple_processes_with_timeout_exception_raised(cache_location):
+    num_of_processes = 10
+    sleep_interval = 1
+    with pytest.raises(Exception):
```
rayluo, line 86: The direction of last 6 commits were probably not the way we want to go. Let's try go back to the state of right after the [7th-to-last commit](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/58/commits/93a55513febc2a09c8b083fbc1da1dbb060cec6a), and then:

* Remove this `raises(Exception)`
* Refactor `_validate_result_in_cache()` to return the count
* At the end of this test case, `assert count < num_of_process * 2`

And we will go from there.

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,90 @@
...
+            data = cache_accessor.load()
+            if data is None:
+                data = ""
+            data += "< " + str(os.getpid()) + "\n"
+            time.sleep(sleep_interval)
+            data += "> " + str(os.getpid()) + "\n"
+            cache_accessor.save(data)
+    except Exception as e:
+        logging.warning("Exception raised %s", e)
+        sys.exit(3)
```
rayluo, line 54:
```suggestion
```

Let's try rollback this function to the previous one using NO exception logic

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,90 @@
...
+            args=(cache_location, sleep_interval))
+        processes.append(t)
+
+    for process in processes:
+        process.start()
+
+    for process in processes:
+        process.join()
+        if process.exitcode == 3:
+            raise Exception
```
rayluo, line 72:
```suggestion
```

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,90 @@
...
+                assert process_id == prev_process_id, "Process overlap found"
+                assert tag == '>', "Process overlap_found"
+                prev_process_id = None
+            else:
+                assert tag == '<', "Opening bracket not found"
+                prev_process_id = process_id
+
+    assert count <= expected_entry_count * 2, "File content corrupted"
+    if count < expected_entry_count * 2:
+        logging.warning("Starvation detected")
```
rayluo, line 37:
```suggestion
    return count
```

rayluo, line 37:
```suggestion
    return count
```

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,90 @@
...
+    _validate_result_in_cache(num_of_processes, cache_location)
+    os.remove(cache_location)
+
+
+def test_multiple_processes_with_timeout_exception_raised(cache_location):
+    num_of_processes = 10
+    sleep_interval = 1
+    with pytest.raises(Exception):
+        _run_multiple_processes(
+            num_of_processes, cache_location, sleep_interval)
```
rayluo, line 88:
```suggestion
        _run_multiple_processes(
            num_of_processes, cache_location, sleep_interval)
        count = _validate(...)
        assert count < num_of_processes * 2, "Should observe starvation"
```

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,83 @@
...
+    for process in processes:
+        process.join()
+
+
+def test_multiple_processes_without_timeout_exception(cache_location):
+    num_of_processes = 20
+    sleep_interval = 0
+    _run_multiple_processes(num_of_processes, cache_location, sleep_interval)
+    count = _validate_result_in_cache(num_of_processes, cache_location)
+    assert count == num_of_processes * 2
```
rayluo: We tried this earlier in this PR. The proper way to do this is to use:

```suggestion
    if count < num_of_process * 2:
        logging.warning("Starvation detected")
```

abhidnya13: Yes, but I am seeing in this case it is less than no of processes because of the locking exception and not starvation, if you check [build outputs](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/58/commits/e19a0b176ef65e54e53b45469e5b1dd45d2bdae8) for this you can verify that,  I was referring to that in our offline teams conversation as well. so if we don't have a timeout, it will work fine.

abhidnya13: So I introduced that logging of exception in the previous 6 commits to see why the test that is supposed to be running for processes without a timeout is failing because of timeout, if it is then we will have to reduce the num of process even more just to make sure the test actually tests what it is supposed to and does not fail because of timeout

rayluo: You called it "timeout (during lock acquisition)", I called it "starvation". Probably neither is a good name, because naming is hard in computer science. :-)

But the point here is, unable to acquire lock during high load is totally understandable and even expected. That is not an error, and we do NOT reduce the number of processes for that. The goal of this particular test case is NOT to avoid such "abortion due to no lock available" (oops I'm inventing new names again), it is to observe whether garbled data would occur when lots of processes accessing the same cache.

Whether/how to detect "abortion due to no lock available" is a different topic. You are welcome to still attempt it, based on all the understandings above.

abhidnya13: I think it does make sense to rename the test names to distinguish the purpose because in both the cases we are ok having a timeout, its just testing them on high load or low load and finding the behaviour only. This tests will have no relation to testing of locks then as that behaviour does not even kick in case of or library because of timeouts.

abhidnya13: Having said that we can always log the timeout exception to see what hapenned there. So I might anyway add that and then may be we add one more test case which succeeds without any difficulty under normal load

rayluo: > we can always log the timeout exception to see what hapenned there. So I might anyway add that

No problem. Just make sure you specifically catch the exception raised by portalocker. Do not just `except Exception`.

> may be we add one more test case which succeeds without any difficulty under normal load
That is also welcome. Make sure we will have 3 meaningful names. :-)

rayluo: > This tests will have no relation to testing of locks then as that behaviour does not even kick in case of or library because of timeouts.

Not quite. When that "timeout (during lock acquisition)" happens, it is exactly the locking mechanism doing its job. And, even when in usual cases when everything seems just working fine, that is still the locking mechanism doing its job on guarding race conditions.

abhidnya13: Sure, that is correct, what I meant is that the performance tests that Petro found issues with was with a really long timeout and then checking how to the lock behaves and thats where we found a few issues on Mac. These tests are definitely not testing that behavior.

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,94 @@
...
+from portalocker import exceptions
+import pytest
+
+from msal_extensions import FilePersistence, CrossPlatLock
+
+
+@pytest.fixture
+def cache_location():
+    path_to_script = os.path.dirname(os.path.abspath(__file__))
+    return os.path.join(path_to_script, str(os.getpid())+"msal.cache")
```
rayluo, line 15: 1. Taking a close look, does the `os.getpid()` work as expected? Wouldn't it always have same PID among our 2+ test cases here, thus unintentionally share same file among all the test cases? While that does not currently seem to make a difference in terms of test output, the 3 test cases still unnecessarily couple with each other.

   Perhaps you can take a look into Python's built-in tempfile module.

2. Also, let's follow `test_persistence.py` approach to close the test file by this fixture, otherwise there seems no point to use the fixture feature to begin with.

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,94 @@
...
+        with CrossPlatLock(lock_file_path):
+            data = cache_accessor.load()
+            if data is None:
+                data = ""
+            data += "< " + str(os.getpid()) + "\n"
+            time.sleep(sleep_interval)
+            data += "> " + str(os.getpid()) + "\n"
+            cache_accessor.save(data)
+    except exceptions.LockException as e:
+        logging.warning("Timeout occured %s", e)
```
rayluo, line 50:
```suggestion
        logging.warning("Unable to acquire lock %s", e)
```

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,94 @@
...
+    num_of_processes = 4
+    sleep_interval = 0.1
+    _run_multiple_processes(num_of_processes, cache_location, sleep_interval)
+    count = _validate_result_in_cache(cache_location)
+    os.remove(cache_location)
+    assert count == num_of_processes * 2, "Should not observe starvation"
+
+
+def test_lock_for_high_workload(cache_location):
+    num_of_processes = 20
```
rayluo, line 79:
```suggestion
    num_of_processes = 80
```

tests/test_cache_lock_file_perf.py
```diff
@@ -0,0 +1,94 @@
...
+    sleep_interval = 0
+    _run_multiple_processes(num_of_processes, cache_location, sleep_interval)
+    count = _validate_result_in_cache(cache_location)
+    os.remove(cache_location)
+    assert count <= num_of_processes * 2, "Should observe starvation"
```
rayluo, line 84:
```suggestion
    assert count <= num_of_processes * 2, "Starvation or not, we should not observe garbled payload"
```

tests/test_cache_lock_file_perf.py
abhidnya13: I can just use the temp_file fixture we have used in test_persistence.py, that seems to be working fine as well. Will update it in the next commit

bgavrilMS:
The goal is to check that our extensions work side by side. First manually, ideally automated. I think having a console app / script will help this. 

I won't be able to look at this work item for a while though, we should discuss if someone else can pick it up.

tests/cache_file_generator.py
```diff
@@ -1,3 +1,11 @@
+"""
+This is a console application which is to be used for cross-platform lock performance testing.
+The app will acquire lock for the cache file, log the process id and free the lock
```
rayluo, line 3:
```suggestion
The app will acquire lock for the cache file, log the process id and release the lock
```

I believe the jargon is acquire and release.

By the way, since you go this far, you might as well do [this](https://stackoverflow.com/questions/7791574/how-can-i-print-a-python-files-docstring-when-executing-it), to also help this CLI tool's user.

abhidnya13: Nice suggestion, love the idea mentioned in the link, making that change.

PR #59, opened 2020-05-19
Update setup.py

abhidnya13:


abhidnya13:
Thanks for the detailed explanation 👍  I think it makes sense, will close this PR

PR #60, opened 2020-06-03
Cache file console app

abhidnya13:
Adding console app for cross-platform testing in #60

### DIFF
--- a/tests/cache_file_generator.py
+++ b/tests/cache_file_generator.py
@@ -1,6 +1,28 @@
+import logging
+import os
 import sys
+import time
 
-from test_cache_lock_file_perf import _acquire_lock_and_write_to_cache
+from portalocker import exceptions
+
+from msal_extensions import FilePersistence, CrossPlatLock
+
+
+def _acquire_lock_and_write_to_cache(cache_location, sleep_interval):
+    cache_accessor = FilePersistence(cache_location)
+    lock_file_path = cache_accessor.get_location() + ".lockfile"
+    try:
+        with CrossPlatLock(lock_file_path):
+            data = cache_accessor.load()
+            if data is None:
+                data = ""
+            data += "< " + str(os.getpid()) + "\n"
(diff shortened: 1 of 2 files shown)


tests/cache_file_generator.py
```diff
@@ -0,0 +1,6 @@
+import sys
+
+from test_cache_lock_file_perf import _acquire_lock_and_write_to_cache
```
rayluo, line 3: I would suggest to do it the other way around:
* Move `_acquire_lock_and_write_to_cache()` from that test runner to here
* Let the test runner import this helper function.

Not only because this would be more logically correct, it would also remove one unnecessary dependency - the `pytest` - when .Net team is going to use this generator as a stand-alone script.

abhidnya13: Hey Ray, you are right, this makes sense, I have made that change in my latest commit !

tests/cache_file_generator.py
```diff
@@ -0,0 +1,29 @@
...
+import os
+import sys
+import time
+
+from portalocker import exceptions
+
+from msal_extensions import FilePersistence, CrossPlatLock
+
+
+def _acquire_lock_and_write_to_cache(cache_location, sleep_interval):
```
rayluo, line 11: I believe @bgavrilMS wants:

> accepts 3 inputs: string protectedFilePath, string lockFilePath and int millisecondsToSleep

abhidnya13: I can change to take this parameter as an input but I was just wondering if its actually needed as it follows the same pattern across libraries

rayluo: Fair enough, unless @bgavrilMS has any objection. :-)

Now, how about we just have this new `cache_file_console_app` branch merge back to your `perf_testing` branch, and then we conclude this whole thing there?

bgavrilMS: No, this is fine.

PR #63, opened 2020-07-28
Cache resiliency

abhidnya13:


msal_extensions/cache_lock.py
```diff
@@ -25,7 +28,24 @@ def __init__(self, lockfile_path):
             flags=portalocker.LOCK_EX | portalocker.LOCK_NB,
             **open_kwargs)
 
+    def try_to_create_lock_file(self):
+        retries_no = 100
+        retry_delay_milliseconds = 0.150
+        for i in range(retries_no):
+            try:
+                open(self._lockpath, 'x')
```
rayluo, line 36: Do we need to wrap this inside a `with ...` clause so that the open file handler would be closed before it being opened (again) later? It might now really matter, but feels the right thing to do...

msal_extensions/cache_lock.py
```diff
@@ -25,7 +28,24 @@ def __init__(self, lockfile_path):
...
+    def try_to_create_lock_file(self):
+        retries_no = 100
+        retry_delay_milliseconds = 0.150
+        for i in range(retries_no):
+            try:
+                with open(self._lockpath, 'x'):
+                    return True
+            except OSError as err:
+                if err.errno == errno.EEXIST:
+                    time.sleep(retry_delay_milliseconds)
```
rayluo, line 40: [Isn't this](https://docs.python.org/3/library/time.html#time.sleep) `retry_delay_in_secs`, rather than `milliseconds`?

msal_extensions/cache_lock.py
```diff
@@ -25,7 +28,24 @@ def __init__(self, lockfile_path):
...
 
+    def try_to_create_lock_file(self):
+        retries_no = 100
+        retry_delay_milliseconds = 0.150
+        for i in range(retries_no):
+            try:
+                with open(self._lockpath, 'x'):
+                    return True
+            except OSError as err:
+                if err.errno == errno.EEXIST:
```
rayluo, line 39: What do we want to do with the `else` clause? Is the current missing `else` intended?

msal_extensions/cache_lock.py
```diff
@@ -25,7 +28,24 @@ def __init__(self, lockfile_path):
             flags=portalocker.LOCK_EX | portalocker.LOCK_NB,
             **open_kwargs)
 
+    def try_to_create_lock_file(self):
+        retries_no = 100
+        retry_delay_milliseconds = 0.150
```
rayluo, line 33: Do we have a reasoning for choosing this pair of numbers? What did our Java friends use here?
FWIW, portalocker's default settings is 0.25 interval, and 5 seconds waiting maximum. (BTW, I also like the way they control this by maximum waiting time, rather than by retries_no.)

rayluo (commented):
Overall structure looks good. Now let's dive into the details.

msal_extensions/cache_lock.py
abhidnya13: I need to verify these values from Java team. I dont know if these are good defaults. I have pinged Petro to go through the PR but I guess I can ask him what defaults have been used in Java. From what I recollect its 6 secs waiting time not sure what the interval is. I agree we can control this by maximum waiting time rather than retries_no. I will make that change after verifying the  default timeout values.I think portalocker's default settings look reasonable to me.

msal_extensions/cache_lock.py
abhidnya13: Good catch, will make the change!

msal_extensions/cache_lock.py
```diff
@@ -39,7 +39,10 @@ def try_to_create_lock_file(self):
                     return True
             except OSError as err:
                 if err.errno == errno.EEXIST:
+                    logging.warning("Lock file exists, trying again after some time")
```
rayluo, line 42: Long story short, please change all `logging` into `logger`, after we follow [this pattern](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/0.3.0/msal_extensions/libsecret.py#L16-L18).

rayluo (approved):
Overall looks good. Your latest commit reminds me that I missed out an issue in earlier commits. :-)  Please refer to the comment below. I'm still leaving an approval here in advance.

msal_extensions/cache_lock.py
```diff
@@ -25,7 +32,29 @@ def __init__(self, lockfile_path):
...
+                    return True
+            except OSError as err:
+                if err.errno == errno.EEXIST:
+                    logger.warning("Lock file exists, trying again after some time")
+                    time.sleep(check_interval)
+                else:
+                    raise
+            except ValueError:
+                logger.warning("Python 2 does not support atomic creation of file")
+                return False
```
rayluo, line 51: On second thought, since we are already excluding Python 2 anyway, we do not have to use all those `OSError`, `error.EEXIST`, and the tricky `else: raise`. The following might just work. Mind give it a shot?

```suggestion
            except ValueError:  # This needs to be the first clause, for Python 2 to hit it
                logger.warning("Python 2 does not support atomic creation of file")
                return False
            except FileExistsError:  # Only Python 3 will reach this clause
                logger.warning("Lock file exists, trying again after some time")
                time.sleep(check_interval)
```

msal_extensions/cache_lock.py
```diff
@@ -25,7 +32,29 @@ def __init__(self, lockfile_path):
             flags=portalocker.LOCK_EX | portalocker.LOCK_NB,
             **open_kwargs)
 
+    def try_to_create_lock_file(self):
+        timeout = 5
+        check_interval = 0.25
```
rayluo, line 37: Also, let's move that module-level `current_time()` definition to here.
```suggestion
        check_interval = 0.25
        current_time = getattr(time, "monotonic", time.time)
```

rayluo (approved):
Good job! Ship it! :-)

PR #65, opened 2020-07-30
Handling OS errors raised if input is empty and unrecognized for the first usage 

### LINKED ISSUE #64: Investigate OS errors returned for encryption scenarios on Windows and MacOS
Cache file previously used with MSAL already exists before using MSAL Extensions for the very first time. 

For macOS:

Keychain Error: ITEM_NOT_FOUND was returned as service name and account name entry is not found 

For windows:

If the file has no data: WinError 87: The parameter is incorrect -  is returned.
If the file has cache stored in previous MSAL cache format(not encrypted) - returns WinError 13- The data is invalid 

----------------------------------------------------------------------------------------------------------------------------------

There is another WinError that is currently raised at this point : OSError: [WinError -2146893813] : '' for which the cause is unknown. Have not been able to reproduce it yet. Adding the stack trace here:
```C:\Users\kinshu>bonsai configure -w afaf
Traceback (most recent call last):
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\kinshu\AppData\Local\Programs\Python\Python37\Scripts\bonsai.exe\__main__.py", line 7, in <module>
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\bonsai_cli\bonsai.py", line 3457, in main
    cli()
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\click-7.0-py3.7.egg\click\core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\click-7.0-py3.7.egg\click\core.py", line 717, in main
    rv = self.invoke(ctx)
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\click-7.0-py3.7.egg\click\core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\click-7.0-py3.7.egg\click\core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\click-7.0-py3.7.egg\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\click-7.0-py3.7.egg\click\decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\bonsai_cli\bonsai.py", line 3254, in configure
    bonsai_config = Config(use_aad=True, require_workspace=False)
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\bonsai_cli\config.py", line 215, in __init__
    self.accesskey = self.aad_client.get_access_token()
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\bonsai_cli\aad.py", line 109, in get_access_token
    token = self._get_access_token_from_cache()
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\bonsai_cli\aad.py", line 96, in _get_access_token_from_cache
    accounts = self._app.get_accounts()
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\msal-1.0.0-py3.7.egg\msal\application.py", line 308, in get_accounts
    accounts = self._find_msal_accounts(environment=self.authority.instance)
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\msal-1.0.0-py3.7.egg\msal\application.py", line 328, in _find_msal_accounts
    TokenCache.CredentialType.ACCOUNT, query={"environment": environment})
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\msal_extensions\token_cache.py", line 55, in find
    self._reload_if_necessary()
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\msal_extensions\token_cache.py", line 36, in _reload_if_necessary
    self.deserialize(self._persistence.load())
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\msal_extensions\persistence.py", line 124, in load
  File "c:\users\kinshu\appdata\local\programs\python\python37\lib\site-packages\msal_extensions\windows.py", line 111, in unprotect
    raise OSError(256, '', '', err_code)
OSError: [WinError -2146893813] : '' ```

### PR DESCRIPTION
**Pull request by abhidnya13**
Fixes #64 
For Keychain:
Scenario:
- Cache file already exists before using MSAL extensions for the very first time.
- Gives an error as cannot find corresponding service name and account name entry in Keychain 
- Returned error is KeychainError : ITEM_NOT_FOUND

For Windows:
- Empty cache file already exists before using MSAL extensions 
- Adds a check to not make call to DPAPI if input is an empty file.
- Existing code returns an error which is OS error: WinError 87

-----------------------------------------------------------------------------------------------------------------------
Added on 8/4/2020: [Wiki](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/Migrate-MSAL-Python-Cache-to-MSAL-Extensions) page summarizing migration scenario from MSAL Python to MSAL Extensions using same file.

### DIFF
--- a/msal_extensions/osx.py
+++ b/msal_extensions/osx.py
@@ -172,6 +172,7 @@ class Keychain(object):
             contents,
             None,
         )
+        
         if exit_status:
             raise KeychainError(exit_status=exit_status)
 
--- a/msal_extensions/osx.py
+++ b/msal_extensions/osx.py
@@ -172,7 +172,7 @@ class Keychain(object):
             contents,
             None,
         )
-        
+
         if exit_status:
             raise KeychainError(exit_status=exit_status)
 
--- a/msal_extensions/osx.py
+++ b/msal_extensions/osx.py
@@ -172,6 +172,8 @@ class Keychain(object):
             contents,
             None,
         )
+        if exit_status == Keychain.ITEM_NOT_FOUND:
+            return None
 
         if exit_status:
             raise KeychainError(exit_status=exit_status)
--- a/msal_extensions/osx.py
+++ b/msal_extensions/osx.py
@@ -172,9 +172,6 @@ class Keychain(object):
             contents,
             None,
         )
-        if exit_status == KeychainError.ITEM_NOT_FOUND:
-            return None
-
         if exit_status:
             raise KeychainError(exit_status=exit_status)
 
--- a/msal_extensions/osx.py
+++ b/msal_extensions/osx.py
@@ -172,7 +172,7 @@ class Keychain(object):
             contents,
             None,
         )
-        if exit_status == Keychain.ITEM_NOT_FOUND:
+        if exit_status == KeychainError.ITEM_NOT_FOUND:
             return None
 
         if exit_status:
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -14,7 +14,6 @@ try:
 except:
     from pathlib2 import Path  # An extra lib for Python 2
 
-
 try:
     ABC = abc.ABC
 except AttributeError:  # Python 2.7, abc exists, but not ABC
@@ -153,8 +152,13 @@ class KeychainPersistence(BasePersistence):
 
     def load(self):
         with self._Keychain() as locker:
-            return locker.get_generic_password(
-                self._service_name, self._account_name)
+            from .osx import KeychainError
+            try:
+                return locker.get_generic_password(
+                    self._service_name, self._account_name)
+            except KeychainError as kce:
+                if kce.exit_status != KeychainError.ITEM_NOT_FOUND:
+                    raise
 
     def time_last_modified(self):
         return self._file_persistence.time_last_modified()
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -119,7 +119,6 @@ class FilePersistenceWithDataProtection(FilePersistence):
             handle.write(self._dp_agent.protect(content))
 
     def load(self):
-        """Return decrypted content, or None if input is empty."""
         # type: () -> str
         with open(self._location, 'rb') as handle:
             return self._dp_agent.unprotect(handle.read())
@@ -150,7 +149,6 @@ class KeychainPersistence(BasePersistence):
         self._file_persistence.touch()  # For time_last_modified()
 
     def load(self):
-        """Return content from OSX Keychain, or None if the content does not exist."""
         with self._Keychain() as locker:
             return locker.get_generic_password(
                 self._service_name, self._account_name)
@@ -190,7 +188,6 @@ class LibsecretPersistence(BasePersistence):
             self._file_persistence.touch()  # For time_last_modified()
 
     def load(self):
-        """Loads content in the secret service, return None when found nothing"""
         return self._agent.load()
 
     def time_last_modified(self):
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -119,12 +119,11 @@ class FilePersistenceWithDataProtection(FilePersistence):
             handle.write(self._dp_agent.protect(content))
 
     def load(self):
-        # type: () -> str
+        """Return decrypted content, or None if input is empty."""
+        # type: () -> Optional[str]
         with open(self._location, 'rb') as handle:
             data = handle.read()
-            if data is not b"":
-                return self._dp_agent.unprotect(data)
-            return None
+            return self._dp_agent.unprotect(data) if data else None
 
 
 class KeychainPersistence(BasePersistence):
@@ -152,14 +151,16 @@ class KeychainPersistence(BasePersistence):
         self._file_persistence.touch()  # For time_last_modified()
 
     def load(self):
+        """Return content from OSX Keychain, or None if the content does not exist."""
+        from .osx import KeychainError
         with self._Keychain() as locker:
-            from .osx import KeychainError
             try:
                 return locker.get_generic_password(
                     self._service_name, self._account_name)
             except KeychainError as kce:
-                if kce.exit_status != KeychainError.ITEM_NOT_FOUND:
-                    raise
+                if kce.exit_status == KeychainError.ITEM_NOT_FOUND:
+                    return None  # Map ITEM_NOT_FOUND exception to None return value
+                raise
 
     def time_last_modified(self):
         return self._file_persistence.time_last_modified()
@@ -196,6 +197,7 @@ class LibsecretPersistence(BasePersistence):
             self._file_persistence.touch()  # For time_last_modified()
 
     def load(self):
+        """Loads content in the secret service, return None when found nothing"""
         return self._agent.load()
 
     def time_last_modified(self):
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -123,7 +123,7 @@ class FilePersistenceWithDataProtection(FilePersistence):
         # type: () -> Optional[str]
         with open(self._location, 'rb') as handle:
             data = handle.read()
-            return self._dp_agent.unprotect(data) if data else None
+        return self._dp_agent.unprotect(data) if data else None
 
 
 class KeychainPersistence(BasePersistence):
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -124,7 +124,7 @@ class FilePersistenceWithDataProtection(FilePersistence):
             data = handle.read()
             if data is not b"":
                 return self._dp_agent.unprotect(data)
-        return data
+            return None
 
 
 class KeychainPersistence(BasePersistence):
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -14,6 +14,7 @@ try:
 except:
     from pathlib2 import Path  # An extra lib for Python 2
 
+
 try:
     ABC = abc.ABC
 except AttributeError:  # Python 2.7, abc exists, but not ABC
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -120,10 +120,9 @@ class FilePersistenceWithDataProtection(FilePersistence):
 
     def load(self):
         """Return decrypted content, or None if input is empty."""
-        # type: () -> Optional[str]
+        # type: () -> str
         with open(self._location, 'rb') as handle:
-            data = handle.read()
-        return self._dp_agent.unprotect(data) if data else None
+            return self._dp_agent.unprotect(handle.read())
 
 
 class KeychainPersistence(BasePersistence):
@@ -152,15 +151,9 @@ class KeychainPersistence(BasePersistence):
 
     def load(self):
         """Return content from OSX Keychain, or None if the content does not exist."""
-        from .osx import KeychainError
         with self._Keychain() as locker:
-            try:
-                return locker.get_generic_password(
-                    self._service_name, self._account_name)
-            except KeychainError as kce:
-                if kce.exit_status == KeychainError.ITEM_NOT_FOUND:
-                    return None  # Map ITEM_NOT_FOUND exception to None return value
-                raise
+            return locker.get_generic_password(
+                self._service_name, self._account_name)
 
     def time_last_modified(self):
         return self._file_persistence.time_last_modified()
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -121,7 +121,10 @@ class FilePersistenceWithDataProtection(FilePersistence):
     def load(self):
         # type: () -> str
         with open(self._location, 'rb') as handle:
-            return self._dp_agent.unprotect(handle.read())
+            data = handle.read()
+            if data is not b"":
+                return self._dp_agent.unprotect(data)
+        return data
 
 
 class KeychainPersistence(BasePersistence):
--- a/msal_extensions/token_cache.py
+++ b/msal_extensions/token_cache.py
@@ -42,8 +42,9 @@ class PersistedTokenCache(msal.SerializableTokenCache):
                     from .osx import KeychainError
                     if exp.exit_status != KeychainError.ITEM_NOT_FOUND:
                         raise
-                raise
-            # Otherwise, from cache's perspective, a nonexistent file is a NO-OP
+                else:
+                    raise
+                    # Otherwise, from cache's perspective, a nonexistent file is a NO-OP
 
     def modify(self, credential_type, old_entry, new_key_value_pairs=None):
         with CrossPlatLock(self._lock_location):
--- a/msal_extensions/token_cache.py
+++ b/msal_extensions/token_cache.py
@@ -1,5 +1,6 @@
 """Generic functions and types for working with a TokenCache that is not platform specific."""
 import os
+import sys
 import warnings
 import time
 import errno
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by abhidnya13** [comment]
Hey Ray, 

Thanks for reviewing ! 

1) Added the description to the issue
2) Yes non empty cache file would cause the issue, I have added this in description, this fix however would not solve that error. 
Also, after testing, using serialize of token_cache_previous and save in new extensions persistence in the same file(the migration scenario we talked about) does not work either. So the best way to handle this would be to either empty the already existing cache file or use a new one. 
3) For having it in persistence layer, I did think  about that option, however since this is an macOS specific error, I think having it is osx.py makes more sense. Also, we already do such kind of handling in the same file [here](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/dev/msal_extensions/osx.py#L218)

File: msal_extensions/osx.py [hunk 1]
```diff
@@ -172,6 +172,8 @@ def get_generic_password(self, service, account_name):
             contents,
             None,
         )
+        if exit_status == KeychainError.ITEM_NOT_FOUND:
+            return None
```
rayluo, line 176: > For having it in persistence layer, I did think about that option, however since this is an macOS specific error, I think having it is osx.py makes more sense

The problem of "fixing" it here in osx.py is that, the introduction of a new `None` return value would unnecessarily modify/broaden the meaning of this method. `None` was not an expected outcome of this method, per its original design. And it seems we do not have to change that design here.

Would be great to see an alternative of fixing this in persistence layer and see how it works.

msal_extensions/persistence.py
```diff
@@ -150,8 +153,13 @@ def save(self, content):
 
     def load(self):
         with self._Keychain() as locker:
-            return locker.get_generic_password(
-                self._service_name, self._account_name)
+            from .osx import KeychainError
```
rayluo, line 156: Sorry I was not always precise, but I mean to say [(at the very end of here)](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/65#pullrequestreview-458826870) to fix it in the token_cache layer.

And then even those new test cases would probably need to be moved to a new test_token_cache.py instead.

abhidnya13: Hey @rayluo , 
I understand your concern to add it in the token_cache layer but we can sync up offline on the complications involved in moving it there. Since this is a OS specific error, and token_cache layer is shared across OS, there will eventually be import errors for platforms that are not macOS as they wouldnt recognize the except statement and fail. I still think it should just be added in osx.py like it previously was. But, I do understand your concern so  adding it in persistence layer would be still better than token_cache.py.

msal_extensions/persistence.py
```diff
@@ -121,7 +121,10 @@ def save(self, content):
     def load(self):
```
rayluo, line 121: Given that we now head to the direction of using None as an indication of empty input, we better make such intention explicit.

```suggestion
    def load(self):
        """Return decrypted content, or None if input is empty."""
```

msal_extensions/persistence.py
```diff
@@ -121,7 +121,10 @@ def save(self, content):
...
         # type: () -> str
         with open(self._location, 'rb') as handle:
-            return self._dp_agent.unprotect(handle.read())
+            data = handle.read()
+            if data is not b"":
```
rayluo, line 125:
```suggestion
            if data:
```

Alternatively, if you are in a geeky mood:

> return self._dp_agent.unprotect(data) if data else None

msal_extensions/persistence.py
```diff
@@ -150,8 +153,13 @@ def save(self, content):
 
     def load(self):
```
rayluo, line 154: Given that we now head to the direction of using `None` as an indication of ITEM_NOT_FOUND, we better make such intention explicit.
```suggestion
    def load(self):
        """Return content from OSX Keychain, or None if the content does not exist."""
```

msal_extensions/persistence.py
rayluo: Let's move this import to before the `with ...` statement.

msal_extensions/persistence.py
```diff
@@ -150,8 +153,13 @@ def save(self, content):
...
         with self._Keychain() as locker:
-            return locker.get_generic_password(
-                self._service_name, self._account_name)
+            from .osx import KeychainError
+            try:
+                return locker.get_generic_password(
+                    self._service_name, self._account_name)
+            except KeychainError as kce:
+                if kce.exit_status != KeychainError.ITEM_NOT_FOUND:
+                    raise
```
rayluo, line 162: Given that we now head to the direction of using `None` as an indication of ITEM_NOT_FOUND, we better make such intention explicit.

```suggestion
                if kce.exit_status == KeychainError.ITEM_NOT_FOUND:
                    return None  # Map ITEM_NOT_FOUND exception to None return value
                raise
```

tests/test_macos_backend.py
```diff
@@ -43,3 +50,18 @@ def test_osx_token_cache_roundtrip():
...
+@pytest.mark.skipif(
+    is_running_on_travis_ci, reason="Requires manual testing")
+def test_macos_no_keychain_entry_exists_before_first_use():
+    test_folder = tempfile.mkdtemp(prefix="msal_extension_test_windows_token_cache_roundtrip")
+    cache_file = os.path.join(test_folder, 'msal.cache')
+    open(cache_file, 'w+')
+    try:
+        # Make sure key chain entry does not already exist for below service name
+        persistence = KeychainPersistence(cache_file, "my_service_name", "my_account_name")
+        app = msal.PublicClientApplication(
```
rayluo, line 63: Let's move these test cases into test_persistence.py

Also, is it possible to have these 2 new test cases NOT depend on msal.PublicClientApplication?

msal_extensions/persistence.py
abhidnya13: Added in [this](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/65/commits/c4a9938c2ce187bea8767fee3f05b7d8680f9010) commit

msal_extensions/persistence.py
abhidnya13: Added in [this](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/65/commits/c4a9938c2ce187bea8767fee3f05b7d8680f9010) commit

msal_extensions/persistence.py
```diff
@@ -119,9 +119,11 @@ def save(self, content):
             handle.write(self._dp_agent.protect(content))
 
     def load(self):
-        # type: () -> str
+        """Return decrypted content, or None if input is empty."""
+        # type: () -> Optional[str]
         with open(self._location, 'rb') as handle:
-            return self._dp_agent.unprotect(handle.read())
+            data = handle.read()
+            return self._dp_agent.unprotect(data) if data else None
```
rayluo, line 126: We should have been doing this, but we did not. Since we now have an explicit `data` anyway, let's do it the right way then.
```suggestion
        return self._dp_agent.unprotect(data) if data else None
```

tests/test_persistence.py
```diff
@@ -52,3 +52,44 @@ def test_libsecret_persistence(temp_location):
         {"my_attr_1": "foo", "my_attr_2": "bar"},
         ))
 
+@pytest.mark.skipif(
+    is_running_on_travis_ci, reason="Requires manual testing")
+def test_windows_empty_file_exists_before_first_use():
+    test_folder = tempfile.mkdtemp(prefix="msal_extension_test_windows_token_cache_roundtrip")
```
rayluo, line 58: Unlike other existing test files, the test_persistence.py already uses the `temp_location` approach which seems working well. Would recommend to use it here.

tests/test_persistence.py
```diff
@@ -52,3 +52,44 @@ def test_libsecret_persistence(temp_location):
...
+    cache_file = os.path.join(test_folder, 'msal.cache')
+    open(cache_file, 'w+')
```
rayluo, line 60:
```suggestion
    open(cache_file, 'w')
```
By the way, this conceptually also be inside the try...finally... block, unless we switch to use the `temp_location` approach, as suggested above.

tests/test_persistence.py
```diff
@@ -52,3 +52,44 @@ def test_libsecret_persistence(temp_location):
...
+    try:
+        persistence = FilePersistenceWithDataProtection(cache_file)
+        assert persistence.load() is None  # Empty file should just be a NO-OP
+    finally:
+        shutil.rmtree(test_folder, ignore_errors=True)
+
+@pytest.mark.skipif(
+    is_running_on_travis_ci, reason="Requires manual testing")
+def test_windows_file_contains_data_before_first_use():
```
rayluo, line 69: I do not really think this test case is necessary, because this is the default and expected behavior.

If you really want to add it, that's also fine, but please follow the suggestion above.

tests/test_persistence.py
```diff
@@ -52,3 +52,44 @@ def test_libsecret_persistence(temp_location):
...
+    try:
+        persistence = FilePersistenceWithDataProtection(cache_file)
+        with pytest.raises(OSError) as err:
+            persistence.load()
+        assert err.value.winerror == 13  # WinError 13 - The data is invalid
+    finally:
+        shutil.rmtree(test_folder, ignore_errors=True)
+
+@pytest.mark.skipif(
+    is_running_on_travis_ci, reason="Requires manual testing")
```
rayluo, line 85: Have you run these test cases locally? Wouldn't it always fail one way or another because they ALL try to run regardless whether you are on Windows or Mac?

tests/test_persistence.py
```diff
@@ -52,3 +52,44 @@ def test_libsecret_persistence(temp_location):
...
+def test_macos_no_keychain_entry_exists_before_first_use():
+    test_folder = tempfile.mkdtemp(prefix="msal_extension_test_windows_token_cache_roundtrip")
+    cache_file = os.path.join(test_folder, 'msal.cache')
+    open(cache_file, 'w+')
```
rayluo, line 89: Same suggestion earlier.

tests/test_persistence.py
```diff
@@ -52,3 +52,44 @@ def test_libsecret_persistence(temp_location):
...
+    try:
+        # Make sure key chain entry does not already exist for below service name
+        persistence = KeychainPersistence(cache_file, "my_service_name", "my_account_name")
```
rayluo, line 92:
```suggestion
        persistence = KeychainPersistence(
            cache_file, "nonexist_service_name", "nonexist_account_name")
```

tests/test_persistence.py
abhidnya13: I just have to run them separately not as a file, so basically just run the windows one on a windows and macOS ones on Mac

abhidnya13: But if you try to run ALL of them at once it some cases will fail depending on the platform

abhidnya13: Added `skipif` condition to make sure the tests are skipped based on platform in [this](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/65/commits/fab3c72576419c8a839a352f99e18b489109f297) commit

msal_extensions/persistence.py
```diff
@@ -119,9 +119,11 @@ def save(self, content):
             handle.write(self._dp_agent.protect(content))
 
     def load(self):
-        # type: () -> str
+        """Return decrypted content, or None if input is empty."""
+        # type: () -> Optional[str]
         with open(self._location, 'rb') as handle:
-            return self._dp_agent.unprotect(handle.read())
+            data = handle.read()
+        return self._dp_agent.unprotect(data) if data else None
```
bgavrilMS, line 126: If data is None,  it means the cache is corrupt. Because even if you log out all the users, the cache file should not be empty. So you can handle this here, but I would also handle the more general case of "cache is corrupt" (i.e. MSAL fails to parse the cache) and delete the cache file / keychain item / keyring entry

msal_extensions/persistence.py
```diff
@@ -149,9 +151,16 @@ def save(self, content):
...
+        from .osx import KeychainError
         with self._Keychain() as locker:
-            return locker.get_generic_password(
-                self._service_name, self._account_name)
+            try:
+                return locker.get_generic_password(
+                    self._service_name, self._account_name)
+            except KeychainError as kce:
+                if kce.exit_status == KeychainError.ITEM_NOT_FOUND:
+                    return None  # Map ITEM_NOT_FOUND exception to None return value
```
bgavrilMS, line 162: Yes, this is part of the happy path scenario. If you install AzCLI on Mac, you have no cache. MSAL will try to read it, but there is nothing there.

msal_extensions/persistence.py
abhidnya13: Agreed, it will be good to have a general recovery method in the case of cache being corrupt. Created an [issue](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/issues/68) to track this.

msal_extensions/persistence.py
```diff
@@ -119,9 +119,11 @@ def save(self, content):
             handle.write(self._dp_agent.protect(content))
 
     def load(self):
-        # type: () -> str
+        """Return decrypted content, or None if input is empty."""
+        # type: () -> Optional[str]
         with open(self._location, 'rb') as handle:
-            return self._dp_agent.unprotect(handle.read())
+            data = handle.read()
```
henrik-me, line 125: >data = handle.read() [](start = 12, length = 20)

if data is empty then return as empty file

msal_extensions/persistence.py
henrik-me, line 126: >None [](start = 59, length = 4)

per discussion, the API currently would expect this to be empty string

msal_extensions/token_cache.py
```diff
@@ -37,8 +38,13 @@ def _reload_if_necessary(self):
                 self._last_sync = time.time()
         except IOError as exp:
             if exp.errno != errno.ENOENT:
-                raise
-            # Otherwise, from cache's perspective, a nonexistent file is a NO-OP
+                if sys.platform.startswith('darwin'):
+                    from .osx import KeychainError
+                    if exp.exit_status != KeychainError.ITEM_NOT_FOUND:
+                        raise
+                else:
```
rayluo, line 45: This implementation does not address similar exceptions from other platforms, does it?

And, I have a feeling that, even after we complete all those platform-dependent logic here, this would not look as good as the #67 approach.

Feel free to explore more, and then we discuss.

abhidnya13: I remember closing the conversation on the fact that we just need to address the Keychain error for now and so for the time being we do it in the token cache layer. Similar errors in other platforms can be addressed by creating a new Exception at the persistence layer later. I remember the reason we do not want to introduce a None return value is because persistence layer is exposed as an API and so introducing a new exception at that layer is also something we possibly need to discuss. So eventually once we have had that conversation we can go ahead and visit #67. @henrik-me Am I missing something ?

PR #67, opened 2020-08-08
Unified exception when load() is called before save()

### LINKED ISSUE #64: Investigate OS errors returned for encryption scenarios on Windows and MacOS

rayluo:
Hi Abhi, while pondering #65 , I feel uncomfortable of its introduction of a new `None` return value, and today I also realize that such a new behavior in the other 3 persistence flavours are inconsistent with `FilePersistence`. So, in this PR, I'm experimenting a different approach, to NOT introduce the `None` as a successful return value, but rather raise exception - this time we raise a common `IOError`. Test cases are not yet added. Please review this proof-of-concept.

UPDATE at 8/28: We settled with a new exception `PersistenceNotFound` for the scenario described in #64 
This PR will replace/close #65, and resolve #64.

### DIFF
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -50,8 +50,9 @@ def _mkdir_p(path):
 # otherwise caller would need to catch os-specific persistence exceptions.
 class PersistenceNotFound(OSError):
     def __init__(
-            self, err_no=errno.ENOENT, message="Persistence not found", where=None):
-        super(PersistenceNotFound, self).__init__(err_no, message, where)
+            self,
+            err_no=errno.ENOENT, message="Persistence not found", location=None):
+        super(PersistenceNotFound, self).__init__(err_no, message, location)
 
 
 class BasePersistence(ABC):
@@ -115,7 +116,7 @@ class FilePersistence(BasePersistence):
                     message=(
                         "Persistence not initialized. "
                         "You can recover by calling a save() first."),
-                    where=self._location,
+                    location=self._location,
                     )
             raise
 
@@ -129,7 +130,7 @@ class FilePersistence(BasePersistence):
                     message=(
                         "Persistence not initialized. "
                         "You can recover by calling a save() first."),
-                    where=self._location,
+                    location=self._location,
                     )
             raise
 
@@ -171,7 +172,7 @@ class FilePersistenceWithDataProtection(FilePersistence):
                     message=(
                         "Persistence not initialized. "
                         "You can recover by calling a save() first."),
-                    where=self._location,
+                    location=self._location,
                     )
             logger.exception(
                 "DPAPI error likely caused by file content not previously encrypted. "
@@ -214,7 +215,8 @@ class KeychainPersistence(BasePersistence):
                     # This happens when a load() is called before a save().
                     # We map it into cross-platform error for unified catching.
                     raise PersistenceNotFound(
-                        where="%s %s".format(self._service_name, self._account_name),
+                        location="Service:{} Account:{}".format(
+                            self._service_name, self._account_name),
                         message=(
                             "Keychain persistence not initialized. "
                             "You can recover by call a save() first."),
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -9,6 +9,7 @@ app developer would naturally know whether the data are protected by encryption.
 import abc
 import os
 import errno
+import logging
 try:
     from pathlib import Path  # Built-in in Python 3
 except:
@@ -19,6 +20,9 @@ try:
     ABC = abc.ABC
 except AttributeError:  # Python 2.7, abc exists, but not ABC
     ABC = abc.ABCMeta("ABC", (object,), {"__slots__": ()})  # type: ignore
+
+
+logger = logging.getLogger(__name__)
 
 
 def _mkdir_p(path):
@@ -41,6 +45,15 @@ def _mkdir_p(path):
             raise
 
 
+# We do not aim to wrap every os-specific exception.
+# Here we define only the most common one,
+# otherwise caller would need to catch os-specific persistence exceptions.
+class PersistenceNotFound(OSError):
+    def __init__(
+            self, err_no=errno.ENOENT, message="Persistence not found", where=None):
+        super(PersistenceNotFound, self).__init__(err_no, message, where)
+
+
 class BasePersistence(ABC):
     """An abstract persistence defining the common interface of this family"""
 
@@ -57,7 +70,7 @@ class BasePersistence(ABC):
         # type: () -> str
         """Load content from this persistence.
 
-        Could raise IOError if no save() was called before.
+        Could raise PersistenceNotFound if no save() was called before.
         """
         raise NotImplementedError
 
@@ -65,7 +78,7 @@ class BasePersistence(ABC):
     def time_last_modified(self):
         """Get the last time when this persistence has been modified.
 
-        Could raise IOError if no save() was called before.
+        Could raise PersistenceNotFound if no save() was called before.
         """
         raise NotImplementedError
 
@@ -93,11 +106,32 @@ class FilePersistence(BasePersistence):
     def load(self):
         # type: () -> str
         """Load content from this persistence"""
-        with open(self._location, 'r') as handle:
-            return handle.read()
+        try:
+            with open(self._location, 'r') as handle:
+                return handle.read()
+        except EnvironmentError as exp:  # EnvironmentError in Py 2.7 works across platform
+            if exp.errno == errno.ENOENT:
+                raise PersistenceNotFound(
+                    message=(
+                        "Persistence not initialized. "
+                        "You can recover by calling a save() first."),
+                    where=self._location,
+                    )
+            raise
+
 
     def time_last_modified(self):
-        return os.path.getmtime(self._location)
+        try:
+            return os.path.getmtime(self._location)
+        except EnvironmentError as exp:  # EnvironmentError in Py 2.7 works across platform
+            if exp.errno == errno.ENOENT:
+                raise PersistenceNotFound(
+                    message=(
+                        "Persistence not initialized. "
+                        "You can recover by calling a save() first."),
+                    where=self._location,
+                    )
+            raise
 
     def touch(self):
         """To touch this file-based persistence without writing content into it"""
@@ -127,9 +161,22 @@ class FilePersistenceWithDataProtection(FilePersistence):
 
     def load(self):
         # type: () -> str
-        with open(self._location, 'rb') as handle:
-            data = handle.read()
-        return self._dp_agent.unprotect(data)
+        try:
+            with open(self._location, 'rb') as handle:
+                data = handle.read()
+            return self._dp_agent.unprotect(data)
+        except EnvironmentError as exp:  # EnvironmentError in Py 2.7 works across platform
+            if exp.errno == errno.ENOENT:
+                raise PersistenceNotFound(
+                    message=(
+                        "Persistence not initialized. "
+                        "You can recover by calling a save() first."),
+                    where=self._location,
+                    )
+            logger.exception(
+                "DPAPI error likely caused by file content not previously encrypted. "
+                "App developer should migrate by calling save(plaintext) first.")
+            raise
 
 
 class KeychainPersistence(BasePersistence):
@@ -165,9 +212,13 @@ class KeychainPersistence(BasePersistence):
             except self._KeychainError as ex:
                 if ex.exit_status == self._KeychainError.ITEM_NOT_FOUND:
                     # This happens when a load() is called before a save().
-                    # We map it to cross-platform IOError for unified catching.
-                    raise IOError(errno.ENOENT, "Content is empty", "%s %s".format(
-                        self._service_name, self._account_name))
+                    # We map it into cross-platform error for unified catching.
+                    raise PersistenceNotFound(
+                        where="%s %s".format(self._service_name, self._account_name),
+                        message=(
+                            "Keychain persistence not initialized. "
+                            "You can recover by call a save() first."),
+                        )
                 raise  # We do not intend to hide any other underlying exceptions
 
     def time_last_modified(self):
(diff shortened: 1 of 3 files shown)


abhidnya13:
Hey Ray, Thanks for getting this perspective, love to see different ways of doing things ! 
I went through the implementation but I am still not sure why do you feel uncomfortable returning a None value. I would love get more perspective on that. I think we should rather have a conversation before coming up with test cases.

rayluo:
Good question! The reason I feel uncomfortable of `None` is still because it is not a valid outcome in the first place. Bogdan happens to also voice his similar concern [here](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/65#discussion_r468086602). Feel free to ping me in Teams.

msal_extensions/persistence.py
```diff
@@ -41,6 +45,15 @@ def _mkdir_p(path):
             raise
 
 
+# We do not aim to wrap every os-specific exception.
+# Here we define only the most common one,
+# otherwise caller would need to catch os-specific persistence exceptions.
+class PersistenceNotFound(OSError):
+    def __init__(
+            self, err_no=errno.ENOENT, message="Persistence not found", where=None):
```
abhidnya13, line 53: `where` seems a little confusing to me , can we use something like `location` or `persistence_location`?

msal_extensions/persistence.py
```diff
@@ -87,11 +106,32 @@ def save(self, content):
...
         """Load content from this persistence"""
-        with open(self._location, 'r') as handle:
-            return handle.read()
+        try:
+            with open(self._location, 'r') as handle:
+                return handle.read()
+        except EnvironmentError as exp:  # EnvironmentError in Py 2.7 works across platform
+            if exp.errno == errno.ENOENT:
+                raise PersistenceNotFound(
+                    message=(
```
abhidnya13, line 115: I think we really shouldn't be encouraging the use of this persistence layer as a public API. Although we already have Azure CLI using it for one of their use cases, we don't really expect people to be using this load() and save() functions for storing msal cache. In such a case we should make sure this persistence not found error is never raised to anyone using msal extensions with the token_cache layer  only.

msal_extensions/persistence.py
```diff
@@ -41,6 +45,15 @@ def _mkdir_p(path):
             raise
 
 
+# We do not aim to wrap every os-specific exception.
+# Here we define only the most common one,
+# otherwise caller would need to catch os-specific persistence exceptions.
+class PersistenceNotFound(OSError):
```
abhidnya13, line 51: I am still thinking about `PersistenceNotFound` as an appropriate name of this error. Especially because right now it only caters to file not being found error. If there is a need in the future to address anything other than file not found, then this name would cause an issue. How about having a generic `PersistenceError`  instead ?

msal_extensions/persistence.py
```diff
@@ -115,13 +155,28 @@ def __init__(self, location, entropy=''):
 
     def save(self, content):
         # type: (str) -> None
+        data = self._dp_agent.protect(content)
```
abhidnya13, line 158: Can you remind me why we need this change ?

msal_extensions/persistence.py
```diff
@@ -115,13 +155,28 @@ def __init__(self, location, entropy=''):
...
+            return self._dp_agent.unprotect(data)
+        except EnvironmentError as exp:  # EnvironmentError in Py 2.7 works across platform
+            if exp.errno == errno.ENOENT:
+                raise PersistenceNotFound(
+                    message=(
+                        "Persistence not initialized. "
+                        "You can recover by calling a save() first."),
+                    where=self._location,
+                    )
+            logger.exception(
```
abhidnya13, line 176: We know this can be one of the reasons but such a logging without explicity checking it might be misleading for cases where a different dpapi issue is causing unprotect to fail. We can either just skip this hardcoded message or if we want to log what is causing the error, we can use the a message from the caught exception instead

msal_extensions/persistence.py
rayluo: After reconsideration, I still think `PersistenceNotFound` (or perhaps we can change it to `PersistenceNotExist`) is the right meaning we want to convey. The current name was chosen, NOT because we "cater to" the underlying FileNotFound error, but because that is the right semantics the higher level wants, i.e. our cache layer can confidently treat a "Persistence Not Found" as "cache is empty; business as usual".

Should we have a need in the future to address anything other than the underlying file-not-found, they are unlikely able to be safely ignored. So, they would NOT go into this error anyway.

By the way, that was also why the leading comments in line 48 says, "we do not aim to wrap everything" into this exception. Therefore, a more generic `PersistenceError` is NOT what we want here.

msal_extensions/persistence.py
rayluo: Admittedly, `where` is a weird name. The reason it was not named as `location` or `persistence_location` was that, in the macOS scenario, it is a combination of service_name and account_name, which I am not sure whether they fit into "location" concept. Maybe they do. Let me know. I am open for any suggestion here.

msal_extensions/persistence.py
rayluo: We discussed this offline, that this persistence layer is mean to be part of the public API. So, no actionable item on this part, for now. We could, however, consider to even make this persistence layer a new library by itself. But that is a different topic that we can discuss offline.

msal_extensions/persistence.py
rayluo: Not strictly needed. We were just splitting old line 119 into new line 158 and 160.

msal_extensions/persistence.py
rayluo: Will it help if we highlight the keyword "LIKELY" at line 177? This message does not mean to be a decisive conclusion, but we are giving a most likely reason and then provide a self-troubleshooting suggestion.  #supportability

msal_extensions/persistence.py
```diff
@@ -87,11 +106,32 @@ def save(self, content):
...
+            raise
+
 
     def time_last_modified(self):
-        return os.path.getmtime(self._location)
+        try:
+            return os.path.getmtime(self._location)
+        except EnvironmentError as exp:  # EnvironmentError in Py 2.7 works across platform
+            if exp.errno == errno.ENOENT:
+                raise PersistenceNotFound(
```
rayluo, line 128: One more thing. Now that we start to work on this PR, @abhidnya13 , would you mind backport your test cases from #65 to this PR?

msal_extensions/persistence.py
abhidnya13: I was thinking of a case where the persistence is corrupted and we want to handle such a case. Or even the case where file is empty but it does exist. I am concerned of the cases where we are expected to recover from or would be a no-op. I guess my question is why do we not aim to wrap everything in this exception ? So that we dont need a new one if any more cases need to be handled. Also we can always check this at the token_cache layer with an err_no like we did before to make is a no-op.

msal_extensions/persistence.py
abhidnya13: @rayluo You can go ahead and backport the test cases, I think they still might need some changes and cannot be used as is. Right now it the tests are of the token cache layer

msal_extensions/persistence.py
rayluo: Short answer is, if the persistence is corrupted, this kind of error will and should be bubbled up as a different exception, which will unlikely be treated as no-op. We literally do NOT aim to wrap everything here. This exception is just one "leaf" of a future exception hierarchy.

abhidnya13: But if we are to bubble it up as a different exception later, will it be ok adding such an exception later as persistence layer is public  API ?

msal_extensions/persistence.py
rayluo: After discussion, we agreed on `location`. Will change in next commit.

msal_extensions/persistence.py
rayluo: After discussion, we agreed it is OK to only have this `PersistenceNotFound` here, and it will be OK in the future if we wrap more "not previously-catchable" underlying platform specific exceptions into a meaningful new exception.

msal_extensions/persistence.py
rayluo: Test cases added. But automation only covers few platforms. Please help manually test the Windows platform, and optionally the mac. I'll manually test on Linux.

abhidnya13: Manually tested the windows platform and mac and the tests for those are green so looks good !

rayluo: Manual test on Linux also passed.

```
(env37) demo@antix1:~/Live-usb-storage/repo/microsoft-authentication-extensions-for-python (handle-error-of-load-before-save)$ pytest tests/test_persistence.py -v                                                                                                             
============================================================================================================================= test session starts ==============================================================================================================================
platform linux -- Python 3.7.3, pytest-5.2.1, py-1.8.0, pluggy-0.13.0 -- /home/demo/Live-usb-storage/env37/bin/python3
cachedir: .pytest_cache
rootdir: /home/demo/Live-usb-storage/repo/microsoft-authentication-extensions-for-python
collected 8 items                                                                                                                                                                                                                                                              

tests/test_persistence.py::test_file_persistence PASSED                                                                                                                                                                                                                  [ 12%]
tests/test_persistence.py::test_nonexistent_file_persistence PASSED                                                                                                                                                                                                      [ 25%]
tests/test_persistence.py::test_file_persistence_with_data_protection SKIPPED                                                                                                                                                                                            [ 37%]
tests/test_persistence.py::test_nonexistent_file_persistence_with_data_protection SKIPPED                                                                                                                                                                                [ 50%]
tests/test_persistence.py::test_keychain_persistence SKIPPED                                                                                                                                                                                                             [ 62%]
tests/test_persistence.py::test_nonexistent_keychain_persistence SKIPPED                                                                                                                                                                                                 [ 75%]
tests/test_persistence.py::test_libsecret_persistence PASSED                                                                                                                                                                                                             [ 87%]
tests/test_persistence.py::test_nonexistent_libsecret_persistence PASSED  <=== Manual test passed
```

PR #69, opened 2020-08-14
Handle file not found exceptions on Python 2.7

chlowell:
On Python 3, calling `PersistedTokenCache.find()` on a nonexistent cache file is a no-op, so this code works:
```py
from msal_extensions import FilePersistence, PersistedTokenCache

persistence = FilePersistence('nope')
cache = PersistedTokenCache(persistence=persistence)
cache.find('')
```
On Python 2.7 it raises a platform-specific error: `WindowsError` on Windows, `OSError` on Linux (I haven't tried it on macOS). The difference in behavior is due to `PersistedTokenCache` anticipating `IOError` when a file isn't found. Neither `WindowsError` nor `OSError` is a subclass of `IOError`. All these errors are subclasses of `EnvironmentError` though, and that class has the `errno` attribute, so a simple fix is to handle `EnvironmentError` here.

### DIFF
--- a/msal_extensions/token_cache.py
+++ b/msal_extensions/token_cache.py
@@ -35,7 +35,7 @@ class PersistedTokenCache(msal.SerializableTokenCache):
             if self._last_sync < self._persistence.time_last_modified():
                 self.deserialize(self._persistence.load())
                 self._last_sync = time.time()
-        except IOError as exp:
+        except EnvironmentError as exp:
             if exp.errno != errno.ENOENT:
                 raise
             # Otherwise, from cache's perspective, a nonexistent file is a NO-OP
--- a/tests/test_agnostic_backend.py
+++ b/tests/test_agnostic_backend.py
@@ -44,3 +44,8 @@ def test_current_platform_cache_roundtrip_with_alias_class(temp_location):
 def test_persisted_token_cache(temp_location):
     _test_token_cache_roundtrip(PersistedTokenCache(FilePersistence(temp_location)))
 
+def test_file_not_found_error_is_not_raised():
+    persistence = FilePersistence('non_existing_file')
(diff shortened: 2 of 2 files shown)


rayluo:
Thanks for the finding! How urgent is this? We have been working on some relevant PRs here, and could cut a release within a week or two. Will that work?

chlowell:
There are workarounds but it affects our core getting started scenario, so one week is much better than two. Will the release be 0.2.3?

rayluo:
Well then, let's aim for a release with this fix at no later than next Monday (that is one week).

PR wise, it would be better to see some test case(s) to guard a potential future regression. If Charles can provide some, it would be great; otherwise @abhidnya13 and I will still pick this up within next couple days.

abhidnya13:
Thanks @chlowell for the code snippet. I created a [branch](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/tree/file_not_found_error_testing) to add a test case for this.
We can see that - 
- [**BEFORE**](https://travis-ci.org/github/AzureAD/microsoft-authentication-extensions-for-python/builds/720806883) state - where the test fails for Python 2.7 ONLY. Our build in travis ci does not test Python 2.7 on Mac but I tested this on my Mac with Python 2.7 and we get the same `OSError` as in Linux.
- [**AFTER**](https://travis-ci.org/github/AzureAD/microsoft-authentication-extensions-for-python/builds/720808521) state - where we change the code to use `Environment Error` and the test case passes on all the platforms and all the versions. 
@rayluo I can create a PR to be merged in this branch for adding such a test to verify. How do you want to go about this ?

rayluo:
@abhidnya13 Thanks very much for the before-and-after comparison!

Either creating a new PR (and linking back to this one), or adding your test case to this PR. They are both good.

PR #73, opened 2020-09-01
Release 0.3.0

abhidnya13:
- New unified `PersistenceNotFound` exception is now raised for cases where the persistence is not found. (#64, #67)
- Bugfix: File not found exception is now handled for Python 2.7 as a no-op (#69)
- Added performance tests for locking behavior (#58)
- A non-exist persistence on Linux platform would previously return a `None`. Since this release, it will raise `PersistenceNotFound` exception which becomes a consistent behavior on Windows and macOS.


abhidnya13:
Thanks @rayluo. Updating the release notes.

PR #76, opened 2020-09-17
Mention licence

### LINKED ISSUE #75: Add license and SCM to setup.py
Could you please add a license and source code declaration to the setup.py file?

Currently msal-extensions causes a violation in our OSS check as these license and source code location fields are missing for this package on https://pypi.org/project/msal-extensions/.
Not a Python guy but believe this can be fixed by adding the following to https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/dev/setup.py#L16-L35:

```
license='MIT',
classifiers=[
       'Development Status :: 2 - Pre-Alpha',
       'License :: OSI Approved :: MIT License '
],
project_urls={
        'Source': 'https://github.com/AzureAD/microsoft-authentication-extensions-for-python.git'
},
```

rayluo:
Mention license. Also add project_urls which would probably trigger an automatic project home page link being rendered in PyPI. This would fix #75 , although we would only see the result at next release.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,11 @@
+# https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files
+
 [bdist_wheel]
 universal=1
+
+[metadata]
+license = MIT
+project_urls =    Changelog = https://github.com/AzureAD/microsoft-authentication-extensions-for-python/releases
+classifiers =
+    License :: OSI Approved :: MIT License
+    Development Status :: 2 - Pre-Alpha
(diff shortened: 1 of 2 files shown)


abhidnya13 (commented):
Just curious, is there a reason we are doing this in `setup.cfg` and not `setup.py` ? I was referring to MSAL and I see most of this information is in `setup.py` except the changelog url.

rayluo:
Yeah we are tentatively [experimenting the setup.cfg approach](https://stackoverflow.com/a/60885212).

abhidnya13 (approved):
:shipit:

Issue #83, opened 2021-02-17
Token cache persistence fails on Windows if home dir is drive root


mparry:
If you have your home directory on Windows set to be the root of a drive, e.g. `HOME=H:\`, then the token cache persistence fails with an `EACCES` permissions error:

```
  File "lib\site-packages\msal_extensions\token_cache.py", line 75, in __init__
    FilePersistenceWithDataProtection(cache_location, entropy=entropy))
  File "lib\site-packages\msal_extensions\persistence.py", line 159, in __init__
    super(FilePersistenceWithDataProtection, self).__init__(location)
  File "lib\site-packages\msal_extensions\persistence.py", line 103, in __init__
    _mkdir_p(os.path.dirname(self._location))
  File "lib\site-packages\msal_extensions\persistence.py", line 40, in _mkdir_p
    os.makedirs(path)
  File "lib\os.py", line 223, in makedirs
    mkdir(name, mode)
PermissionError: [WinError 5] Access is denied: 'H:\\'
```

The simplest/best solution would be to update the `persistence._mkdir_p()` implementation to look like this instead:

```python
def _mkdir_p(path):
    if path:
        os.makedirs(path, exist_ok=True)
```

However, this requires Python ≥3.2. Considering that Python 2 was EOL at the start of 2020, this seems reasonable to me! Clearly, you might not wish to do so just for such a relatively small issue.

Alternatively, either the code could call the above only if Python is new enough (and so the bug persists under Python 2) or else it would presumably be necessary to implement roughly the same bug fix as they did in Python some years ago: https://bugs.python.org/issue25583

rayluo:
Can you provide the exact input to our API, to help us reproduce this and better understand it? The existing `if not path: return  # NO-OP` is expected to skip the `os.makedirs(path)` already.

```
$ python -c "import os; print(len(os.path.dirname('H:\\foo.txt')))"
0
```

mparry:
Seems to be sufficient to just do something like

```python
from msal_extensions import TokenCache
TokenCache("O:\\foo.txt")
```

And indeed for your example, I don't get 0, I get 3:

```python
> python -c "import os; print(len(os.path.dirname('O:\\foo.txt')))"
3
```

(This is using Python 3.9.1 on Windows 10, FWIW.)

mparry:
The associated PR uses the built-in functionality if Python is new enough, which is certainly one simple solution. Obviously you may prefer to fix it for Python 2 too.

[...]

PR #84, opened 2021-02-18
Fixed failure when a user's home dir is a drive root on Windows

mparry:
Note that the bug remains under Python 2

Addresses #83


mparry:
The build failure looks unrelated

mparry:
@rayluo - this now contains your change too

PR #86, opened 2021-04-07
Call `_reload_if_necessary` in `PersistedTokenCache.__init__`

jiasli:
Currently `_reload_if_necessary` is only lazily called when `modify` or `find` is called.

However, if `modify` or `find` is not called, `SerializableTokenCache.serialize` won't work as expected because `self._cache` is `{}`. The caller must workaround it by either 

1. call `modify` or `find` once
2. manually call the protected method `_reload_if_necessary`

### DIFF
--- a/msal_extensions/token_cache.py
+++ b/msal_extensions/token_cache.py
@@ -26,6 +26,7 @@ class PersistedTokenCache(msal.SerializableTokenCache):
         self._persistence = persistence
         self._last_sync = 0  # _last_sync is a Unixtime
         self.is_encrypted = persistence.is_encrypted
+        self._reload_if_necessary()
 
     def _reload_if_necessary(self):
         # type: () -> None


rayluo:
> Currently `_reload_if_necessary` is only lazily called when `modify` or `find` is called.
> 
> However, if `modify` or `find` is not called, `SerializableTokenCache.serialize` won't work as expected because `self._cache` is `{}`. The caller must workaround it by either
> 
>     1. call `modify` or `find` once
> 
>     2. manually call the protected method `_reload_if_necessary`

The observation is correct. The `PersistedTokenCache` was designed to be used by MSAL, and MSAL only calls `find()` and `modify()`. In fact, `PersistedTokenCache` should have inherited from `TokenCache`, rather than `SerializableTokenCache`. If we implemented it that way, there won't be an issue of "`PersistedTokenCache.serialize()` won't work ...", because `PersistedTokenCache.serialize()` won't even exist.

The proposal in this PR won't benefit MSAL because that extra load at the initialization is unnecessary to MSAL. So, now the topic becomes "why would you want to call `PersistedTokenCache.serialize()` explicitly, especially before any `PersistedTokenCache.find()`". We can discuss and see whether it can be justified to add that behavior into this library.

jiasli:
Previously, we don't have a way to disable token cache. Now we have `core.encrypt_token_cache` option to disable token encryption (https://github.com/Azure/azure-cli/issues/19506), so calling `PersistedTokenCache.serialize` is no loner needed for us.

PR #93, opened 2021-07-22
Raise without log polluting the console

### LINKED ISSUE #92: API for testing whether persistence is available
It would be helpful to have an API for silently testing whether a persistence method is available, in particular LibsecretPersistence. The best way today is to attempt to instantiate the persistence, which logs an error on Linux when PyGObject isn't installed:
```
Runtime dependency of PyGObject is missing.
Depends on your Linux distro, you could install it system-wide by something like:
    sudo apt install python3-gi python3-gi-cairo gir1.2-secret-1
If necessary, please refer to PyGObject's doc:
https://pygobject.readthedocs.io/en/latest/getting_started.html
```
It's a reasonable thing to log but an application developer using msal-extensions through another library may interpret it to mean that mid-tier library requires PyGObject (see e.g. Azure/azure-sdk-for-python/issues/12152 and Azure/azure-sdk-for-python/issues/19857).

rayluo:
This may fix #92 .

@chlowell , continuing our conversation in #92, this PR currently chooses to not use `six` (which is never in MSAL or MSAL EX dependency, by the way).

### DIFF
--- a/msal_extensions/libsecret.py
+++ b/msal_extensions/libsecret.py
@@ -13,33 +13,29 @@ Alternatively, you could skip Cairo & PyCairo, but you still need to do all thes
     pip install wheel
     PYGOBJECT_WITHOUT_PYCAIRO=1 pip install --no-build-isolation pygobject
 """
-import logging
-
-logger = logging.getLogger(__name__)
 
 try:
     import gi  # https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/Encryption-on-Linux  # pylint: disable=line-too-long
 except ImportError:
-    logger.exception(
-        """Runtime dependency of PyGObject is missing.
+    raise ImportError("""Unable to import module 'gi'
+Runtime dependency of PyGObject is missing.
 Depends on your Linux distro, you could install it system-wide by something like:
     sudo apt install python3-gi python3-gi-cairo gir1.2-secret-1
 If necessary, please refer to PyGObject's doc:
 https://pygobject.readthedocs.io/en/latest/getting_started.html
-""")
-    raise
+""")  # Message via exception rather than log
 
 try:
     # pylint: disable=no-name-in-module
     gi.require_version("Secret", "1")  # Would require a package gir1.2-secret-1
     # pylint: disable=wrong-import-position
     from gi.repository import Secret  # Would require a package gir1.2-secret-1
-except (ValueError, ImportError):
-    logger.exception(
+except (ValueError, ImportError) as ex:
+    raise type(ex)(
         """Require a package "gir1.2-secret-1" which could be installed by:
         sudo apt install gir1.2-secret-1
-        """)
-    raise
+        """)  # Message via exception rather than log
+
 
 class LibSecretAgent(object):
     """A loader/saver built on top of low-level libsecret"""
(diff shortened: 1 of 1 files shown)


msal_extensions/libsecret.py
```diff
@@ -134,7 +130,5 @@ def trial_run():
   you may need to install gnome-keyring package.
 * Headless mode (such as in an ssh session) is not supported.
 """
-        logger.exception(message)  # This log contains trace stack for debugging
-        logger.warning(message)  # This is visible by default
-        raise
+        raise RuntimeError(message)  # Message via exception rather than log
```
rayluo, line 133: @chlowell , there are 3 snippets changed in this PR.

* In the first one, the original exception was an `ImportError`, and this PR would also throw an `ImportError`, so that is fine.
* In the second one, the original exception would be `ValueError` or `ImportError`, but it seems good enough for this PR to "normalize" them into - and raise - an `ImportError`.
* The third snippet here, this PR tentatively uses a generic `RuntimeError`. Do you think it would be beneficial to raise a customized `MsalExtensionsRuntimeError` instead?

msal_extensions/libsecret.py
```diff
@@ -13,33 +13,29 @@
...
+""".format(__file__))  # Message via exception rather than log
 
 try:
     # pylint: disable=no-name-in-module
     gi.require_version("Secret", "1")  # Would require a package gir1.2-secret-1
     # pylint: disable=wrong-import-position
     from gi.repository import Secret  # Would require a package gir1.2-secret-1
 except (ValueError, ImportError):
-    logger.exception(
+    raise ImportError(
```
chlowell, line 34: No need to raise a different type:
```py
except (ValueError, ImportError) as ex:
    raise type(ex)("new message")
    # or
    raise ex.__class__("new message")
```

msal_extensions/libsecret.py
```diff
@@ -13,33 +13,29 @@
...
-        """Runtime dependency of PyGObject is missing.
+    raise ImportError("""cannot import name 'gi' from 'libsecret' ({})
+Runtime dependency of PyGObject is missing.
 Depends on your Linux distro, you could install it system-wide by something like:
     sudo apt install python3-gi python3-gi-cairo gir1.2-secret-1
 If necessary, please refer to PyGObject's doc:
 https://pygobject.readthedocs.io/en/latest/getting_started.html
-""")
-    raise
+""".format(__file__))  # Message via exception rather than log
```
chlowell, line 26: `__file__` will already be in the traceback

msal_extensions/libsecret.py
chlowell: I think `RuntimeError` is reasonable, being the built-in type that seems most appropriate to this case. A custom exception is reasonable too but I lean toward the built-in for two reasons:
1. the value of a custom type is in enabling an application to choose a response suited to a specific failure case. This method doesn't know the specific failure case and it's irrelevant anyway because there's nothing a running application can do to make `LibsecretPersistence` work when this method fails.
2. you can change the type later without breaking anyone: `MsalExtensionsRuntimeError(RuntimeError)`

msal_extensions/libsecret.py
rayluo: I was merely trying to mimic the behavior from the underlying `gi` library. It happens to include the file name in the ImportError message, even when it is also available from the traceback.

`ImportError: cannot import name '_gi' from 'gi' (/home/demo/Live-usb-storage/env37/lib/python3.7/site-packages/gi/__init__.py)`

That being said, I agree with you that we don't have to do that. I'll remove the `__file__` thing in my next commit.

PR #95, opened 2021-08-20
Relax `portalocker` pin

### LINKED ISSUE #94: Support `portalocker` >=2
The current constraints restrict `portalocker` to <2:
https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/a0cfe93433400b2d4ea673ebdef4bf064243c13d/setup.py#L25-L26

IIUC the major version bump was just to remove py2 support so this library *should* be compatible with `portalocker>=1.6`

dhirschfeld:
Resolves #94
Resolves https://github.com/conda-forge/msal_extensions-feedstock/issues/11

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,8 +22,8 @@ setup(
     package_data={'': ['LICENSE']},
     install_requires=[
         'msal>=0.4.1,<2.0.0',
-        "portalocker~=1.6;platform_system=='Windows'",
-        "portalocker~=1.0;platform_system!='Windows'",
+        "portalocker>=1.6;platform_system=='Windows'",
+        "portalocker>=1.0;platform_system!='Windows'",
         "pathlib2;python_version<'3.0'",
         ## We choose to NOT define a hard dependency on this.
         # "pygobject>=3,<4;platform_system=='Linux'",
--- a/setup.py
+++ b/setup.py
@@ -22,8 +22,10 @@ setup(
     package_data={'': ['LICENSE']},
     install_requires=[
         'msal>=0.4.1,<2.0.0',
-        "portalocker>=1.6;platform_system=='Windows'",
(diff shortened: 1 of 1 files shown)


dhirschfeld:
Thought I'd see what the CI thought of this change...

dhirschfeld:
So, it looks like this is failing on py27 as it's installing `portalocker-2.3.0`

I'm not a pip expert (as I use `mamba`) but it appears that `portalocker` is missing a [`python-requires`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires):
* https://github.com/WoLpH/portalocker/blob/develop/setup.py#L108-L150

The classifiers imply it works on any Python `>=3.3` but I think the correct way to prevent `pip` from installing version `2.3.0` alongside py27 is to specify `python_requires='>=3.3'`. Failing that restriction in the upstream repo we'd have to specify that here.

Packaging is hard! 🥴

rayluo:
> The classifiers imply it works on any Python `>=3.3`

The classifiers even incorrectly included Python 2, until I send out a PR there, which got merged in 6 hours ago. :-)

> but I think the correct way to prevent `pip` from installing version `2.3.0` alongside py27 is to specify `python_requires='>=3.3'`.

Thanks for the hint. I also send out another PR there.

> Failing that restriction in the upstream repo we'd have to specify that here.

This is also true. And we can probably do it by [Environment Markers](https://www.python.org/dev/peps/pep-0508/#environment-markers). Not sure whether we can chain multiple markers, though:

    "portalocker~=1.6;platform_system=='Windows';python_version<'3'",
    "portalocker>=1.6,<3;platform_system=='Windows';python_version>='3'",



> Packaging is hard!

Can't agree more! 🥴

PR #100, opened 2021-11-12
Skip read lock to improve concurrency

rayluo:
An attempt to bypass the locking behavior when reading token cache. Most of the time, token cache is only serving read operation, so conceptually we will only need to lock once per hour. This would likely improve the concurrency a lot.

@jiasli

### DIFF
--- a/msal_extensions/token_cache.py
+++ b/msal_extensions/token_cache.py
@@ -73,9 +73,21 @@ class PersistedTokenCache(msal.SerializableTokenCache):
             self._last_sync = time.time()
 
     def find(self, credential_type, **kwargs):  # pylint: disable=arguments-differ
-        with CrossPlatLock(self._lock_location):
-            self._reload_if_necessary()
-            return super(PersistedTokenCache, self).find(credential_type, **kwargs)
+        # Use optimistic locking rather than CrossPlatLock(self._lock_location)
+        retry = 3
+        for attempt in range(1, retry + 1):
+            try:
+                self._reload_if_necessary()
+            except Exception:  # pylint: disable=broad-except
+                # Presumably other processes are writing the file, causing dirty read
+                if attempt < retry:
+                    logger.debug("Unable to load token cache file in No. %d attempt", attempt)
+                    time.sleep(0.5)
+                else:
+                    raise  # End of retry. Re-raise the exception as-is.
+            else:  # If reload encountered no error, the data is considered intact
+                return super(PersistedTokenCache, self).find(credential_type, **kwargs)
+        return []  # Not really reachable here. Just to keep pylint happy.
 
 
 class FileTokenCache(PersistedTokenCache):


jiasli:
I think this works great for `find`, but we still need to rethink about `modify`. If 2 processes are writing to he same file, maybe we can let one process wait for a while, instead of making a hard failure with `portalocker.LOCK_NB`?

rayluo:
> I think this works great for `find`, but we still need to rethink about `modify`. If 2 processes are writing to he same file, maybe we can let one process wait for a while, instead of making a hard failure with `portalocker.LOCK_NB`?

True. There is still room for improvement for `modify`, which will probably happen in a different PR.

That being said, the improvement on `find` would still indirectly help `modify`, because now there will be zero reader to content for the lock.

jiasli (approved):

jiasli:
I think we'd better accelerate the fix. I have been tripped by this issue myself several times when executing Azure CLI in parallels.

msal_extensions/token_cache.py
```diff
@@ -50,9 +50,21 @@ def modify(self, credential_type, old_entry, new_key_value_pairs=None):
...
+            try:
+                self._reload_if_necessary()
+            except Exception:  # pylint: disable=broad-except
+                # Presumably other processes are writing the file, causing dirty read
+                if attempt < retry:
+                    logger.debug("Unable to load token cache file in No. %d attempt", attempt)
+                    time.sleep(0.5)
+                else:
+                    raise  # End of retry. Re-raise the exception as-is.
+            else:  # If reload encountered no format error, the data is likely intact
```
jiasli, line 65: Wondering what is "no format error".

rayluo: Good point. Now it is reworded in the latest commit.

jiasli:
I think there is still a very small chance that https://github.com/Azure/azure-cli/issues/20273 can be triggered: It is when the token expires. If multiple instances refreshes at the same time, the permission issue may still happen since there is no retry logic:

https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/2c9690ef88ef77da25f3f9f67639bf1ace236fa4/msal_extensions/token_cache.py#L65-L73

This issue has been bugging Azure CLI for years (https://github.com/Azure/azure-cli/issues/9427).

rayluo:
> I think there is still a very small chance that [Azure/azure-cli#20273](https://github.com/Azure/azure-cli/issues/20273) can be triggered: It is when the token expires. If multiple instances refreshes at the same time,

We consider "multiple app instances encountered token expiration *at the same time*" is less likely happening than "multiple app instances run `az login` *at the same time*". The reason is, multiple `az login` would presumably happening at time T, thus racing with each other. But multiple `az foo bar` job workers would likely run in a slightly different pace. It is like the job time adds in some natural jitter.

Please continue to advocate customers to adopt a pattern of "run one `az login` first, followed by a bunch of `az foo bar` job workers". Or, even better, "run one `az login` and then one `az foo bar` to populate the token cache with a fresh AT for scope foo, and then followed by a bunch of `az foo bar` job workers".

> the permission issue may still happen since there is no retry logic:
> 
> https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/2c9690ef88ef77da25f3f9f67639bf1ace236fa4/msal_extensions/token_cache.py#L65-L73

Retry was considered unnecessary for `persistence.save()`, when/if the `with CrossPlatLock(...)` was successful. FWIW, `with CrossPlatLock(...)` already has its own retry logic.

> 
> This issue has been bugging Azure CLI for years ([Azure/azure-cli#9427](https://github.com/Azure/azure-cli/issues/9427)).

True. In a sense, all these "unable to obtain a write lock" issues mean to be better than a token cache file corruption issue ([Azure/azure-cli#9427](https://github.com/Azure/azure-cli/issues/9427)).

Issue #87, opened 2021-04-14
Provide a default persistence per OS


mparry:
In `0.2.0` the simple `TokenCache` interface was replaced by newer API that now requires the user to write some code along these lines:

```python
def build_persistence(location, fallback_to_plaintext=False):
    """Build a suitable persistence instance based your current OS"""
    if sys.platform.startswith('win'):
        return FilePersistenceWithDataProtection(location)
    if sys.platform.startswith('darwin'):
        return KeychainPersistence(location, "my_service_name", "my_account_name")
    if sys.platform.startswith('linux'):
        try:
            return LibsecretPersistence(
                location,
                schema_name="my_schema_name",
                attributes={"my_attr1": "foo", "my_attr2": "bar"},
                )
        except:  # pylint: disable=bare-except
            if not fallback_to_plaintext:
                raise
            logging.exception("Encryption unavailable. Opting in to plain text.")
    return FilePersistence(location)

cache = PersistedTokenCache(build_persistence(...))
```

(That's taken from the README, of course.)

Is there a reason why this logic needs to be reproduced in all user code that doesn't have more particular requirements? Or, conversely, are there many situation where this wouldn't be good enough?

If not, couldn't the library provide a `get_default_persistence(location)` method, that contains something along the lines of the above? I would be happy to raise a PR for this.

mparry:
Side note: we only recently noticed and updated our code to sync with this, since `DeprecationWarning` doesn't surface by default and you only get something more visible on Linux, where a `RuntimeWarning` is issued about the unprotected cache you're getting from the legacy `TokenCache` interface.

rayluo:
Thanks, @mparry, you raise a good point on the usability. I do not currently have a definite answer. Here is the background behind that change.

[...]

PR #103, opened 2021-12-01
Making all platform-dependent parameters optional

rayluo:
## Changes in this PR

Az CLI uses 2 instances of persistence for different data, stored in different files. @jiasli found it unobvious that he would need to use different `service_name` and `account_name` when running on mac. This PR attempts to hide/handle this detail, by automatically generate those 2 parameters based on the (signal) file location.

## Future ideas that are not (yet?) in this PR

1. Historically, before we introduced all those platform-dependent parameters, there was a unified `TokenCache` interface across all platforms.
2. Later when we added (more and more) platform-dependent parameters, we attempted to keep a similar cross-platform `build_persistence()` helper but it did not really hide all those platform-dependent details, it merely becomes a union of all parameters.
   ```python
   def build_persistence(
       ...,
        entropy='',  # For Windows
        service_name=None, account_name=None,  # For OSX
        schema_name="", attributes=None, fallback_to_plaintext=False,  # For Linux
        **kwargs):  # Extra parameters will be relayed to LibsecretPersistence
   ```
   Therefore, @chlowell [convinced us](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/44#discussion_r407705595) to remove that helper, and rightfully so:
   > I feel the helper doesn't really help me because it doesn't hide the particulars of persistence. To use it, I must learn which options apply on which platforms, and how; which ones are truly optional. Then I either pass all options at once or call the helper conditionally by platform. I prefer to write the conditional code because it would be easier for future generations to understand. At that point I may as well just import and configure persistence classes directly, that's nicely explicit.

3. Now, if this PR would be welcomed and merged, the next step would be reconsidering the introduction of a unified helper. This time, it won't need any platform-dependent parameters. Perhaps even a unified Persistence alias:
   ```python
   # in __init__.py
   if sys.platform.startswith('win'):
       from .persistence import FilePersistenceWithDataProtection as EncryptedPesistence
   elif sys.platform.startswith('darwin'):
       from .persistence import KeyChainPersistnce as EncryptedPesistence
   else:
       from .persistence import LibsecretPersistence as EncryptedPesistence
   ```
   This way, the `EncryptedPersistence` would still be different classes with different signatures on different platforms, but at least their required parameters would be unified so that you could initialize one as `EncryptedPersistence('/path/to/my/file.bin')`. And that would help resolve `https://github.com/AzureAD/microsoft-authentication-extensions-for-python/issues/87`.

@chlowell , @jiasli, thoughts?

UPDATE at 12/7/2021: We decided to reject proposal 3, because that alias would end up with different shapes. We postpone proposal 2 to a future release, after we gain some real-world usage with this PR.

### DIFF
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -9,6 +9,7 @@ app developer would naturally know whether the data are protected by encryption.
 import abc
 import os
 import errno
+import hashlib
 import logging
 import sys
 try:
@@ -49,6 +50,9 @@ def _mkdir_p(path):
             pass
         else:
             raise
+
+def _auto_hash(input_string):
+    return hashlib.sha256(input_string.encode('utf-8')).hexdigest()
 
 
 # We do not aim to wrap every os-specific exception.
@@ -197,19 +201,18 @@ class KeychainPersistence(BasePersistence):
     and protected by native Keychain libraries on OSX"""
     is_encrypted = True
 
-    def __init__(self, signal_location, service_name, account_name):
+    def __init__(self, signal_location, service_name=None, account_name=None):
         """Initialization could fail due to unsatisfied dependency.
 
         :param signal_location: See :func:`persistence.LibsecretPersistence.__init__`
         """
-        if not (service_name and account_name):  # It would hang on OSX
-            raise ValueError("service_name and account_name are required")
         from .osx import Keychain, KeychainError  # pylint: disable=import-outside-toplevel
         self._file_persistence = FilePersistence(signal_location)  # Favor composition
         self._Keychain = Keychain  # pylint: disable=invalid-name
         self._KeychainError = KeychainError  # pylint: disable=invalid-name
-        self._service_name = service_name
-        self._account_name = account_name
+        default_service_name = "msal-extensions"  # This is also our package name
+        self._service_name = service_name or default_service_name
+        self._account_name = account_name or _auto_hash(signal_location)
 
     def save(self, content):
         with self._Keychain() as locker:
@@ -247,7 +250,7 @@ class LibsecretPersistence(BasePersistence):
     and protected by native libsecret libraries on Linux"""
     is_encrypted = True
 
-    def __init__(self, signal_location, schema_name, attributes, **kwargs):
+    def __init__(self, signal_location, schema_name=None, attributes=None, **kwargs):
         """Initialization could fail due to unsatisfied dependency.
 
         :param string signal_location:
@@ -262,7 +265,8 @@ class LibsecretPersistence(BasePersistence):
         from .libsecret import (  # This uncertain import is deferred till runtime
             LibSecretAgent, trial_run)
         trial_run()
-        self._agent = LibSecretAgent(schema_name, attributes, **kwargs)
+        self._agent = LibSecretAgent(
+            schema_name or _auto_hash(signal_location), attributes or {}, **kwargs)
         self._file_persistence = FilePersistence(signal_location)  # Favor composition
 
     def save(self, content):
--- a/sample/persistence_sample.py
+++ b/sample/persistence_sample.py
@@ -10,7 +10,7 @@ def build_persistence(location, fallback_to_plaintext=False):
     if sys.platform.startswith('win'):
         return FilePersistenceWithDataProtection(location)
     if sys.platform.startswith('darwin'):
-        return KeychainPersistence(location, "my_service_name", "my_account_name")
+        return KeychainPersistence(location)
     if sys.platform.startswith('linux'):
         try:
             return LibsecretPersistence(
@@ -21,8 +21,6 @@ def build_persistence(location, fallback_to_plaintext=False):
                 # unless there would frequently be a desktop session and
                 # a remote ssh session being active simultaneously.
                 location,
-                schema_name="my_schema_name",
-                attributes={"my_attr1": "foo", "my_attr2": "bar"},
                 )
         except:  # pylint: disable=bare-except
             if not fallback_to_plaintext:
@@ -31,6 +29,7 @@ def build_persistence(location, fallback_to_plaintext=False):
     return FilePersistence(location)
 
 persistence = build_persistence("storage.bin", fallback_to_plaintext=False)
+print("Type of persistence: {}".format(persistence.__class__.__name__))
 print("Is this persistence encrypted?", persistence.is_encrypted)
 
 data = {  # It can be anything, here we demonstrate an arbitrary json object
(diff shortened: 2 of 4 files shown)


jiasli:
Personally, I won't really fancy the idea of unifying everything to `EncryptedPesistence` if their constructor signatures are different. If I need to use platform-specific features, such as `service_name`, `account_name` for MacOS, I will need additional `if...then...` logic, which defeats the purpose of unified interface.

The original `build_persistence` solution sounds better to me.

msal_extensions/persistence.py
```diff
@@ -197,19 +201,17 @@ class KeychainPersistence(BasePersistence):
...
-        if not (service_name and account_name):  # It would hang on OSX
-            raise ValueError("service_name and account_name are required")
         from .osx import Keychain, KeychainError  # pylint: disable=import-outside-toplevel
         self._file_persistence = FilePersistence(signal_location)  # Favor composition
         self._Keychain = Keychain  # pylint: disable=invalid-name
         self._KeychainError = KeychainError  # pylint: disable=invalid-name
-        self._service_name = service_name
-        self._account_name = account_name
+        self._service_name = service_name or _auto_hash(signal_location)
+        self._account_name = account_name or _auto_hash(signal_location)
```
jiasli, line 214: Do we want to use 

1. different `service_name`s for MSAL token cache and service principal entries?
2. same `service_name` and `account_name` for one `signal_location`?

Perhaps we can default `service_name` to `_auto_hash('msal-extensions-python')`? Just a sudden idea.

jiasli:
Related email discussion: _Naming for service_name and account_name for MacOS token encryption_

msal_extensions/persistence.py
rayluo: > Do we want to use
>
> 1. different `service_name`s for MSAL token cache and service principal entries?

Keep in mind that, in this persistence layer, the class `KeychainPersistence` does not know - and does not need to know - whether the data would be MSAL's token cache or service principal entries.

How would you call and use this `KeychainPersistence`? If you have 2 separate instances as `msal_token_cache = KeychainPersistence("/path/to/msal_token_cache.bin")` and `sp_entries = KeychainPersistence("/path/to/az_cli_sp.bin")`, they will naturally have different `service_name` and different `account_name` after this PR got merged.

> 2. same service_name and account_name for one signal_location?

This PR is already doing that. Same `signal_location` would hash into same `service_name`. Same applies to `account_name`.

> 3. Perhaps we can default `service_name` to `_auto_hash('msal-extensions-python')`?

If your point is to use a fixed `service_name`, we could do that. In that case we won't have to hash. We could just use default service_name as `msal-extensions` (note that I removed the `-python` suffix in your idea, because MSAL EX in other language may still want to potentially share the keychain).

I really do not know which one is better, though: using one fixed default value, or using a dynamical hash based on the `signal_location`.

jiasli: > I really do not know which one is better, though: using one fixed default value, or using a dynamical hash based on the `signal_location`.

This is exactly my point! I don't know if `service_name` should be fixed. Perhaps discuss with @xiangyan99 ? 🤣

Here is `azure-identity`'s code:

https://github.com/Azure/azure-sdk-for-python/blob/47108e9251cd706f48bea359a388b6c8fd161faf/sdk/identity/azure-identity/azure/identity/_persistent_cache.py#L81-L84

```py
    if sys.platform.startswith("darwin"):
        # the cache uses this file's modified timestamp to decide whether to reload
        file_path = os.path.expanduser(os.path.join("~", ".IdentityService", cache_name))
        return msal_extensions.KeychainPersistence(file_path, "Microsoft.Developer.IdentityService", account_name)
```

which `service_name` is fixed to `Microsoft.Developer.IdentityService`.

rayluo: Well then, in the latest amendment, we also use the same `Microsoft.Developer.IdentityService` (for the lack of a better name - naming is hard).

rayluo:
> Personally, I won't really fancy the idea of unifying everything to `EncryptedPesistence` if their constructor signatures are different. If I need to use platform-specific features, such as `service_name`, `account_name` for MacOS, I will need additional `if...then...` logic, which defeats the purpose of unified interface.
> 
> **The original `build_persistence` solution sounds better to me.**

@jiasli , did you mean to say that we can now promote [this updated sample's helper `def build_persistence(location, fallback_to_plaintext=False)`](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/103/files#diff-da1f12ff29935e7e28081ae9203d9d0526a2f8753e03bc63e7a456c6bd1191b1R8-R29) into a formal API?

msal_extensions/persistence.py
```diff
@@ -197,19 +201,18 @@ class KeychainPersistence(BasePersistence):
...
         """
-        if not (service_name and account_name):  # It would hang on OSX
-            raise ValueError("service_name and account_name are required")
         from .osx import Keychain, KeychainError  # pylint: disable=import-outside-toplevel
         self._file_persistence = FilePersistence(signal_location)  # Favor composition
         self._Keychain = Keychain  # pylint: disable=invalid-name
         self._KeychainError = KeychainError  # pylint: disable=invalid-name
-        self._service_name = service_name
-        self._account_name = account_name
+        default_service_name = "Microsoft.Developer.IdentityService"  # Stolen from azure-identity
```
jiasli, line 213: Not sure if MSAL should comply with `azure-identity` which is actually a downstream library of MSAL.

jiasli:
> @jiasli , did you mean to say that we can now promote [this updated sample's helper `def build_persistence(location, fallback_to_plaintext=False)`](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/103/files#diff-da1f12ff29935e7e28081ae9203d9d0526a2f8753e03bc63e7a456c6bd1191b1R8-R29) into a formal API?

I have no strong opinion. Actually, Azure CLI would prefer not to use this `build_persistence`, as we tend to use `azure-cli` as the `service_name`, which is some indeed customization which `build_persistence` can't provide.

msal_extensions/persistence.py
rayluo: I understand that name "Microsoft.Developer.IndentityService" is defined in *downstream* "azure-identity". Originally, I thought we could use any name so we might as well just *backport* the name used by "azure-identity".

But you raised a good point. We should NOT knowingly use any downstream specific service name here, because this open-source library `msal-extensions` could potentially be used by app developers to store any information, and we/they probably do not want to store those data inside a specific downstream component's namespece. I'll send a new commit to use `"msal-extensions"` as the default service name.

jiasli (approved):

jiasli:
Tested with Azure CLI in https://github.com/Azure/azure-cli/pull/20636 and it works perfectly.

PR #104, opened 2021-12-09
Release 0.4.0

rayluo:
* Enhancement: Make all platform-dependent parameters optional (#103)
* Enhancement: Provide `PersistenceEncryptError` and `PersistenceDecryptError`, currently raised when encryption on Windows fails. (PR 108, currently under review)
* Enhancement: The data file will be created with `600` permission when running in Unix-like systems. (PR 107)

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -78,3 +78,32 @@ jobs:
         TOXENV: ${{ matrix.toxenv }}
       run: |
         tox
+
+  cd:
+    needs: ci
+    if: github.event_name == 'push' && (startsWith(github.ref, 'refs/tags') || github.ref == 'refs/heads/master')
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - name: Build a package for release
+      run: |
+        python -m pip install build --user
+        python -m build --sdist --wheel --outdir dist/ .
+    - name: Publish to TestPyPI
+      uses: pypa/gh-action-pypi-publish@v1.4.2
+      if: github.ref == 'refs/heads/master'
+      with:
+        user: __token__
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+    - name: Publish to PyPI
+      if: startsWith(github.ref, 'refs/tags')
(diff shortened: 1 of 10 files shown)


jiasli:
We got a problem here.

`azure-identity` is used in Azure CLI by **some modules owned by service team**, but it requires

https://github.com/Azure/azure-sdk-for-python/blob/4db8247d519b48051eb9f88d1c2c44949b6a45ba/sdk/identity/azure-identity/setup.py#L79

```
"msal-extensions~=0.3.0",
```

Azure CLI can't use `msal-extensions` 0.4.0, as it will cause conflict with `azure-identity`. @xiangyan99

rayluo:
> We got a problem here.
> 
> `azure-identity` is used in Azure CLI by **some modules owned by service team** (though `azure-cli-core` doesn't depend on `azure-identity`), but `azure-identity` requires
> 
> https://github.com/Azure/azure-sdk-for-python/blob/4db8247d519b48051eb9f88d1c2c44949b6a45ba/sdk/identity/azure-identity/setup.py#L79
> 
> ```
> "msal-extensions~=0.3.0",
> ```
> 
> Azure CLI can't use `msal-extensions` 0.4.0, as it will cause conflict with `azure-identity`. @xiangyan99

Good finding, @jiasli !

1. @xiangyan99, the `msal-extensions~=0.3.0` declaration accepts `msal-extensions>=0.3.0,<0.4.0`. That was appropriate at that time, because Semantic Versioning allows potential breaking changes between every `0.x.0` versions. The upcoming msal-extensions 0.4.0 does not contain breaking change, so, azure-identity can simply go ahead and upgrade the dependency to `msal-extensions>=0.3.0,<0.4.0`.

2. But then, @jiasli , we will also need the next version of Azure-Identity with that new widened dependency to be released, in order to let Azure CLI's modules picking up the new Azure-Identity package. So, @xiangyan99, when will next azure-identity be released?

3. MSAL Extensions can still proceed with this 0.4.0 release. It is just that, until No.1 and No.2 above happen, the Azure CLI users would still need to manually patch their environment's msal-extensions by installing it from this github repo's dev branch.

jiasli:
> 1. The upcoming msal-extensions 0.4.0 does not contain breaking change, so, azure-identity can simply go ahead and upgrade the dependency to `msal-extensions>=0.3.0,<0.4.0`.

Should be `msal-extensions>=0.3.0,<0.5.0`.

rayluo:
Abandoning this PR, as we chose to go with 1.0.0 release instead.

PR #108, opened 2022-02-04
Provide catch-able exceptions for 2 dpapi errors

rayluo:
This PR provides user-friendly exception messages for real-world observation: https://github.com/Azure/azure-cli/issues/20231, https://github.com/Azure/azure-cli/issues/20759, and https://github.com/Azure/azure-cli/issues/21010 . The default error messages is not (and can not) be specific to Azure CLI, but Azure CLI can choose to catch the 2 new exceptions `PersistenceEncryptionError` and `PersistenceDecryptionError`, and then emit their own error messages.

CC: @jiasli

### DIFF
--- a/msal_extensions/persistence.py
+++ b/msal_extensions/persistence.py
@@ -56,19 +56,31 @@ def _auto_hash(input_string):
 
 
 # We do not aim to wrap every os-specific exception.
-# Here we define only the most common one,
-# otherwise caller would need to catch os-specific persistence exceptions.
-class PersistenceNotFound(IOError):  # Use IOError rather than OSError as base,
+# Here we standardize only the most common ones,
+# otherwise caller would need to catch os-specific underlying exceptions.
+class PersistenceError(IOError):  # Use IOError rather than OSError as base,
+    """The base exception for persistence."""
         # because historically an IOError was bubbled up and expected.
         # https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/0.2.2/msal_extensions/token_cache.py#L38
         # Now we want to maintain backward compatibility even when using Python 2.x
         # It makes no difference in Python 3.3+ where IOError is an alias of OSError.
+    def __init__(self, err_no=None, message=None, location=None):  # pylint: disable=useless-super-delegation
+        super(PersistenceError, self).__init__(err_no, message, location)
+
+
+class PersistenceNotFound(PersistenceError):
     """This happens when attempting BasePersistence.load() on a non-existent persistence instance"""
     def __init__(self, err_no=None, message=None, location=None):
         super(PersistenceNotFound, self).__init__(
-            err_no or errno.ENOENT,
-            message or "Persistence not found",
-            location)
+            err_no=errno.ENOENT,
+            message=message or "Persistence not found",
+            location=location)
+
+class PersistenceEncryptionError(PersistenceError):
+    """This could be raised by persistence.save()"""
+
+class PersistenceDecryptionError(PersistenceError):
+    """This could be raised by persistence.load()"""
 
 
 class BasePersistence(ABC):
@@ -177,7 +189,13 @@ class FilePersistenceWithDataProtection(FilePersistence):
 
     def save(self, content):
         # type: (str) -> None
-        data = self._dp_agent.protect(content)
+        try:
+            data = self._dp_agent.protect(content)
+        except OSError as exception:
+            raise PersistenceEncryptionError(
+                err_no=getattr(exception, "winerror", None),  # Exists in Python 3 on Windows
+                message="Encryption failed: {}. Consider disable encryption.".format(exception),
+                )
         with os.fdopen(_open(self._location), 'wb+') as handle:
             handle.write(data)
 
@@ -186,7 +204,6 @@ class FilePersistenceWithDataProtection(FilePersistence):
         try:
             with open(self._location, 'rb') as handle:
                 data = handle.read()
-            return self._dp_agent.unprotect(data)
         except EnvironmentError as exp:  # EnvironmentError in Py 2.7 works across platform
             if exp.errno == errno.ENOENT:
                 raise PersistenceNotFound(
@@ -199,6 +216,17 @@ class FilePersistenceWithDataProtection(FilePersistence):
                 "DPAPI error likely caused by file content not previously encrypted. "
                 "App developer should migrate by calling save(plaintext) first.")
             raise
+        try:
+            return self._dp_agent.unprotect(data)
+        except OSError as exception:
+            raise PersistenceDecryptionError(
+                err_no=getattr(exception, "winerror", None),  # Exists in Python 3 on Windows
+                message="Decryption failed: {}. "
+                    "App developer may consider this guidance: "
+                    "https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/PersistenceDecryptionError"  # pylint: disable=line-too-long
+                    .format(exception),
+                location=self._location,
+                )
 
 
 class KeychainPersistence(BasePersistence):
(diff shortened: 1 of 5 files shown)


msal_extensions/persistence.py
```diff
@@ -199,6 +216,14 @@ def load(self):
                 "DPAPI error likely caused by file content not previously encrypted. "
                 "App developer should migrate by calling save(plaintext) first.")
             raise
+        try:
+            return self._dp_agent.unprotect(data)
+        except OSError as exception:
+            raise PersistenceDecryptionError(
+                err_no=getattr(exception, "winerror", None),  # Exists in Python 3 on Windows
+                message="Decryption failed: {}. You may have to delete the file.".format(exception),  # pylint: disable=consider-using-f-string
```
bgavrilMS, line 224: in .NET we delete it ourselves.

msal_extensions/windows.py
```diff
@@ -39,6 +39,15 @@ def raw(self):
         _MEMCPY(blob_buffer, pb_data, cb_data)
         return blob_buffer.raw
 
+_err_description = {
+    # Keys came from real world observation, values came from winerror.h (https://errors)
+    -2146893813: "Key not valid for use in specified state.",
+    -2146892987: "The requested operation cannot be completed. "
+        "The computer must be trusted for delegation and "
+        "the current user account must be configured to allow delegation. "
+        "See also https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/enable-computer-and-user-accounts-to-be-trusted-for-delegation",
```
bgavrilMS, line 48: Great!

bgavrilMS (approved):

msal_extensions/persistence.py
rayluo: Thanks for bringing up this intriguing topic.

The current "you may have to (manually) delete the file" advice was indeed influenced by the actual workaround that we gave to those few Azure CLI's end users who ran into this issue. So, when/if those end users took our advice, they implicitly consented and they knew what they were doing.

But it feels uncomfortable for our implementation to do this silently. The deletion is an irrevocable and potentially destructive operation.

Azure CLI team and us have not quite figured out why end users would run into this issue in the first place. Although the following speculation may or may not be the case for Azure CLI, but, if a downstream app would happen to use same filename for plaintext data and encrypted data, it will then be theoretically possible that a plaintext file predates the decryption attempt. In such case, a smooth recovery plan would be for the app developer to gracefully convert the plaintext data into encrypted data:

```python
encrypted_persistence = FilePersistenceWithDataProtection("my_data.bin")
try:
    valuable_data = encrypted_persistence.load()
except PersistenceDecryptionError:
    valuable_data = FilePersistence("my_data.bin").load()
    encrypted_persistence.save(valuable_data)
```

Or, if the app developer would still prefer, they can catch the `PersistenceDecryptionError` and do the file removal there. The decision is in their hands.

I incline to keep this PR's current behavior i.e. NOT automatically delete the file. We may, however, somehow reword the advice to "App developer or end user may choose to delete the file, or app developer could recover by blah blah blah". The "blah blah blah" part will be lengthy and the audience is the app developer. It is probably fine, as we anticipate the app developer to (eventually) catch this exception and translate it into a more user-friendly message to their end users.

msal_extensions/windows.py
```diff
@@ -39,6 +39,15 @@ def raw(self):
         _MEMCPY(blob_buffer, pb_data, cb_data)
         return blob_buffer.raw
 
+_err_description = {
```
jiasli, line 42: Do we have some knowledge on WinError 0 (https://github.com/Azure/azure-cli/issues/20278)?

msal_extensions/windows.py
```diff
@@ -39,6 +39,15 @@ def raw(self):
...
+    # Keys came from real world observation, values came from winerror.h (https://errors)
```
jiasli, line 43: > https://errors

The URL is incorrect.

msal_extensions/windows.py
jiasli: I don't think people can understand this enigma.

msal_extensions/windows.py
rayluo: The "httpS" one was indeed a typo. I meant to use the "http" one.

And yes, I know it is a Microsoft internal website. Do you know some public website that provide similar information that we can refer to? If not, the `http://errors` is better than nothing. It at least leaves a trace for us to triage customer's reports.

msal_extensions/windows.py
rayluo: > I don't think common users can understand this enigma. They are still non-actionable.

That is arguably true. (TBH, that link was a TL;DR for me. LOL.) But my thoughts were:
* Bogdan found this valuable information that might be useful to those who are determined enough to follow that route, so I would keep it here rather than waste it.
* These paragraphs are for the "inner exception" `OSError`. The "outer exception" `PersistenceEncryptionError` and `PersistenceDecryptionError` both end with a slightly-more-actionable sentence.
* Eventually, Azure CLI may still want to catch the `PersistenceEncryptionError` and `PersistenceDecryptionError`, and convert them into an end-user-friendly sentence.

msal_extensions/windows.py
rayluo: I don't know. The "error 0" is particularly strange. The zero is supposed to mean success.

Regardless, like I [admitted here](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/108/files#r801311953), these error description are still more like troubleshooting trace. I should probably change this PR's title to "Provide catch-able exceptions for all dpapi errors". Eventually Azure CLI might still need to catch them and then tell end user "just turn off encryption by configuring ~/.azure/...", which this MSAL EX package won't provide anyway.

msal_extensions/persistence.py
rayluo: UPDATE: Now the exception message contains a link to this newly-created wiki page https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/PersistenceDecryptionError

msal_extensions/windows.py
```diff
@@ -39,6 +39,15 @@ def raw(self):
         _MEMCPY(blob_buffer, pb_data, cb_data)
         return blob_buffer.raw
 
+_err_description = {
+    # Keys came from real world observation, values came from winerror.h (http://errors)
```
jiasli, line 43: Then I think we should let the reader know this link is Microsoft internal, as the MSAL library and its code is public.

```suggestion
    # Keys came from real world observation, values came from winerror.h (http://errors (Microsoft internal))
```

msal_extensions/persistence.py
```diff
@@ -199,6 +216,17 @@ def load(self):
                 "DPAPI error likely caused by file content not previously encrypted. "
                 "App developer should migrate by calling save(plaintext) first.")
             raise
+        try:
+            return self._dp_agent.unprotect(data)
+        except OSError as exception:
+            raise PersistenceDecryptionError(
+                err_no=getattr(exception, "winerror", None),  # Exists in Python 3 on Windows
+                message="Decryption failed: {}. "  # pylint: disable=consider-using-f-string
```
jiasli, line 224: In our project, we disable all `consider-using-f-string` check as this is really annoying.

msal_extensions/persistence.py
jiasli: Azure CLI guarantees the encrypted file and unencrypted file do _not_ have the same name.

https://github.com/Azure/azure-cli/blob/fb805f4bb3167623107c10ec5f34581a838c8a1c/src/azure-cli-core/azure/cli/core/auth/persistence.py#L38

```py
file_extensions = {True: '.bin', False: '.json'}
...
    location += file_extensions[encrypt]
```

PR #109, opened 2022-02-05
MSAL EX Python 1.0.0

rayluo:
Azure Identity needs to bump its dependency version to `"msal-extensions>=0.3.0,<2"`. @xiangyan99

* New: Add a new platform-independent `build_encrypted_persistence()` API. (#87, #110)
* Remove: Old TokenCache API which has been deprecated for 2 years. (#110)
* Enhancement: Make all platform-dependent parameters optional (#103)
* Enhancement: Provide `PersistenceEncryptError` and `PersistenceDecryptError`, currently raised when encryption on Windows fails. (#108)
* Enhancement: The data file will be created with `600` permission when running in Unix-like systems. (#107)

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -78,3 +78,32 @@ jobs:
         TOXENV: ${{ matrix.toxenv }}
       run: |
         tox
+
+  cd:
+    needs: ci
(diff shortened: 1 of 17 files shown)


rayluo:
We contacted Azure SDK team @xiangyan99 offline. The Azure SDK works fine with this MSAL EX 1.0.0, and Azure Identity will update its dependency to "msal-extensions>=0.3.0,<2" in their next release.

CC: @jiasli

PR #113, opened 2022-04-20
Refine error messages of `PersistenceEncryptionError` and `PersistenceDecryptionError`

jiasli:
Since period is already in https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/e30080b5c1b3e2c801bf371631c3c7afd4cda27a/msal_extensions/windows.py#L44

The error message becomes 

```
Decryption failed: [WinError -2146893813] Key not valid for use in specified state.. App developer may consider this guidance: https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/PersistenceDecryptionError: 'C:\\Users\\xxx\\.azure\\msal_token_cache.bin'
```

### DIFF
--- a/msal_extensions/windows.py
+++ b/msal_extensions/windows.py
@@ -46,7 +46,7 @@ _err_description = {
         "The computer must be trusted for delegation and "
         "the current user account must be configured to allow delegation. "
         "See also https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/enable-computer-and-user-accounts-to-be-trusted-for-delegation",
-    13: "The data is invalid",
+    13: "The data is invalid.",
     }
 
 # This code is modeled from a StackOverflow question, which can be found here:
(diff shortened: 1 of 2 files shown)


msal_extensions/windows.py
```diff
@@ -120,4 +122,4 @@ def unprotect(self, cipher_text):
             finally:
                 _LOCAL_FREE(result.pbData)
         err_code = _GET_LAST_ERROR()
-        raise OSError(None, _err_description.get(err_code), None, err_code)
+        raise OSError(None, _err_description.get(err_code, ''), None, err_code)
```
jiasli, line 125: We should use an empty string `''` as the default value. Otherwise, it becomes:

```
[Errno 0] Decryption failed: [WinError 0] None App developer may consider this guidance: https://github.com/AzureAD/microsoft-authentication-extensions-for-python/wiki/PersistenceDecryptionError: 'C:\\Users\\xxx\\.azure\\msal_token_cache.bin'
```

rayluo (approved):
Thanks for the real-world observation!

Issue #96, opened 2021-08-20
Consider removing `portalocker` dependency


chlowell:
Primarily because it requires `pywin32`, which has been a troublesome (transitive) dependency. See for example these Azure SDK issues:

AssertionError upon importing pywin32 in Azure Function: Azure/azure-sdk-for-python#19989
Can't run `python` after installing pywin32 226: Azure/azure-sdk-for-python#8573
DLL load fails on Python 3.8: Azure/azure-sdk-for-python#8292

rayluo:
Thanks for providing lots of materials here. We will take a look.

Currently, the file locking is the reason we use portalocker, because we do not want to get ourselves into the file locking business. We will need some investigation to see what we can do here.

glaubitz:
Has there been any update on this? Upstream `portalocker` is now at 3.0.0 which conflicts with the version constraint in `setup.py`.

rayluo:
> Has there been any update on this? Upstream `portalocker` is now at 3.0.0 which conflicts with the version constraint in `setup.py`. Since most Linux distributions will always update to the latest upstream version of the packages they maintain, this breaks the `msal-extensions` package.

The PR #117 was a ready-to-test 2 years ago, but it seems there was no strong interest. Now I rebased the PR and further polished it. @glaubitz or everyone interested can briefly test it by:

1. Start from a clean virtual environment and then `pip install https://github.com/AzureAD/microsoft-authentication-extensions-for-python/archive/refs/heads/optional-portalocker.zip`. Run `pip list` to confirm that `portalocker` is not intalled.
2. Run this one-liner to further confirm MSAL-Extensions is using the new file-based lock.

   ```
   python -c "from msal_extensions import CrossPlatLock as L; print(L)"
   <class 'msal_extensions.filelock.CrossPlatLock'>
   ```
3. You can actually use this branch with your MSAL-powered app and see whether it still works normally without portalocker.

PR #117, opened 2022-10-20
Make portalocker optional

rayluo:
This is currently an experimental PR. It will not be merged, not until we receive some test feedback. To beta test this PR, you will need to pull it from its feature branch.

Usage: There is no API change. MSAL EX will automatically utilize portalocker when portalocker is available in current environment, and switch to a new portalocker-less implementation otherwise.

Note:

* A lock mechanism is only used to guard against garbled data due to concurrent read/write. A lock mechanism in itself does not make the higher level application more or less performant. When your app is overloaded, previously (i.e. before this PR) you will see `portalocker.exceptions.LockException` which means the locking mechanism is doing its job. After this PR, under the *same* heavy load, you will see `LockError` instead, which also means the new lock mechanism is functioning.
* ~Test automation currently fails, because keyring daemon no longer works. It used to work even in headless test environment. We will look into it at a later time. But this should not affect or block the beta testing of this PR.~
* UPDATE on Dec 2024: [This PR has been used internally for 2+ years without any issue](https://github.com/Azure/azure-cli/issues/30458#issuecomment-2535324362). Thanks @bebound for testing!

CC: @jiasli

### DIFF
--- a/msal_extensions/filelock.py
+++ b/msal_extensions/filelock.py
@@ -1,0 +1,62 @@
+"""A cross-process lock based on exclusive creation of a given file name"""
+import os
+import sys
+import errno
+import time
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+class LockError(RuntimeError):
+    """It will be raised when unable to obtain a lock"""
+
+
+class CrossPlatLock(object):
+    """This implementation relies only on ``open(..., 'x')``"""
+    def __init__(self, lockfile_path):
+        self._lockpath = lockfile_path
+
+    def __enter__(self):
+        self._create_lock_file('{} {}'.format(
+            os.getpid(),
+            sys.argv[0],
+            ).encode('utf-8'))  # pylint: disable=consider-using-f-string
+        return self
+
+    def _create_lock_file(self, content):
+        timeout = 5
+        check_interval = 0.25
+        current_time = getattr(time, "monotonic", time.time)
+        timeout_end = current_time() + timeout
+        while timeout_end > current_time():
+            try:
+                with open(self._lockpath, 'xb') as lock_file:  # pylint: disable=unspecified-encoding
+                    lock_file.write(content)
+                return None  # Happy path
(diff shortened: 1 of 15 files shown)


.github/workflows/python-package.yml
```diff
@@ -26,8 +26,8 @@ jobs:
             toxenv: "py38"
           - python-version: 3.9
             toxenv: "py39"
-          - python-version: 2.7
-            toxenv: "py27"
+          - python-version: "3.10"
+            toxenv: "py310"
```
jiasli, line 30: Python 3.11 is GA. We are planning to support it in next sprint (https://github.com/Azure/azure-cli/pull/24109).

msal_extensions/filelock.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class LockError(RuntimeError):
+    """It will be raised when unable to obtain a lock"""
+
+
+class CrossPlatLock(object):
```
jiasli, line 16: Well, Azure CLI doesn't really need a such a complex lock at the comment. Previously ADAL didn't have locking mechanism and works mostly fine.

rayluo: Fair point. But if the current PR's non-portalocker lock is harmless, then we might as well just use it.

We are open to the dummy lock idea, too. We may work on it at a later time. It can be done, but it is more work, in terms of needing to expand the current MSAL EX api to allow a "no lock" option. (The current file-based lock, on the contrary, is a drop-in replacement for portalocker, and keeps the MSAL EX api intact.)

If the file-based lock does not work well for Azure CLI on SAW, we will prioritize the dummy lock work accordingly.

rayluo:
See [this comment](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/issues/96#issuecomment-2495425834) on how to test this PR.

ashok672 (approved):
Approving based on the testing done by another team for sometime now.

PR #119, opened 2023-01-09
update osx apis

gattjoe:
Hi Ray,

I took a stab at updating the macOS Security APIs so that they are no longer using the recently deprecated ones.  Once this library is updated, I am hoping the Azure CLI will turn on encryption by default (or as an opt-in).  I'm happy to amend this change as necessary.  Looking forward to your feedback.

Joe

### DIFF
--- a/msal_extensions/osx.py
+++ b/msal_extensions/osx.py
@@ -2,10 +2,20 @@
 
 """Implements a macOS specific TokenCache, and provides auxiliary helper types."""
 
-import os
 import ctypes as _ctypes
+import os
 
 OS_RESULT = _ctypes.c_int32  # pylint: disable=invalid-name
+
+
+class CFString(_ctypes.Structure):
+    # https://developer.apple.com/documentation/_COREfoundation/cfstring
+    pass
+
+
+CFIndex = _ctypes.c_int
+kCFStringEncodingUTF8 = 0x08000100
+CFStringRef = _ctypes.POINTER(CFString)
 
 
 class KeychainError(OSError):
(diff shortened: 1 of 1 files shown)


gattjoe:
@rayluo @bgavrilMS @jiasli I see that a new version of MSAL EX was released. Can someone put eyes on this PR?

bgavrilMS:
Hi @gattjoe - @rayluo and me no longer own public client scenarios. Adding @ashok672 for guidance. 

May I ask what's wrong with the deprecated APIs? Moving to new APIs would require a lot of testing.

gattjoe:
@bgavrilMS @ashok672 

The main background is [this PR](https://github.com/Azure/azure-cli/pull/20636) in the Azure Cli, where it was discussed using the encrypted persistence cache, but that the Apple APIs are legacy.  There was an assertion that the way the Azure Cli was packaged would preclude us from using the newest keychain API, which is not the case.

The main source of Apple keychain API information I used is here: https://developer.apple.com/documentation/technotes/tn3137-on-mac-keychains, where it basically says to use the SecItem keychain API. You can sere [here](https://developer.apple.com/documentation/security/1397763-seckeychainfindinternetpassword) and [here](https://developer.apple.com/documentation/security/1398366-seckeychainaddgenericpassword) how Apple has labeled the keychain APIs that MSAL EX is currently using as deprecated.


There is no difference in behavior with using the new API vs. the old. In reviewing some of your conversations with your peers (on GitHub), I can tell that maintaining compatibility is important for MSAL and by extension MSAL EX. As a result, I spent time to make sure the inputs/outputs of the functions are the same.

I have not spoken to anyone on the Azure Cli team. In the PR referenced above, it seems they are aware that the keychain API is old. I am hoping that if we can update the keychain API with this change, I can perhaps influence them to turn on encryption support with a command line flag before making it the default.

Any feedback is appreciated.

bgavrilMS:
Yes, we know the API is deprecated and Apple doesn't offer support for it. But it works and there is no indication that Apple is removing this support anytime soon.

We tried to move to the SecItem API on the .NET equivalent project, only to find out that it requires a bundle id. So the new APIs are more for apps which are notarized / distributed through the Apple Store. This is not the case of Azure CLI. I may be wrong.

gattjoe:
We wouldn't need a bundle id in this context as we are just using the Python bindings to access the APIs. Maybe you can help me convince @jiasli to turn on **_opt in_** support for encrypting the persistence cache in OS X.

DanielOverdevestPZH:
@jiasli extra security on OSX would be nice

PR #122, opened 2023-11-16
Dropping Python 2.7

rayluo:
The existing test automation were failing due to the infrastructure already dropped Python 2.7 support.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9, 2.7]
+        python-version: [3.7, 3.8, 3.9]
         os: [ubuntu-latest, windows-latest, macos-latest]
         include:
           # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-environment-variables-in-a-matrix
@@ -27,8 +27,6 @@ jobs:
             toxenv: "py38"
           - python-version: 3.9
             toxenv: "py39"
-          - python-version: 2.7
-            toxenv: "py27"
           - python-version: 3.9
             os: ubuntu-latest
             lint: "true"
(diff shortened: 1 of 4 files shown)


.github/workflows/python-package.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9, 2.7]
+        python-version: [3.7, 3.8, 3.9]
```
xiangyan99, line 20: Don't you test py 3.10 & 3.11 (and 3.12 is coming)?

setup.cfg
```diff
@@ -1,12 +1,18 @@
...
 project_urls =    Changelog = https://github.com/AzureAD/microsoft-authentication-extensions-for-python/releases
 classifiers =
     License :: OSI Approved :: MIT License
     Development Status :: 5 - Production/Stable
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
```
xiangyan99, line 17: You need to support 3.11 & 3.12, right?

setup.py
```diff
@@ -29,11 +30,8 @@
         # The following 4 lines use the `and` syntax defined here:
         #       https://www.python.org/dev/peps/pep-0508/#grammar
         "portalocker<3,>=1.0;python_version>='3.5' and platform_system!='Windows'",
```
xiangyan99, line 32: You don't need python_version>='3.5' anymore. It is always true. :)

.github/workflows/python-package.yml
rayluo: I mean to have this PR contain minimal changes needed for dropping Python 2.7, so that your review will not be distracted.

There is already another pending PR that is to add py 3.10, 3.11, 3.12 testing.

xiangyan99 (approved):

PR #125, opened 2023-12-21
Set proper lower bound for `portalocker` dependency, drop `packaging` dependency

akx:
`packaging` was only used as a dependency to check for the version of `portalocker` in use.

This drops that dependency and instead corrects the minimum version of `portalocker` to the one checked here.

It should be safe to depend on a version of `portalocker` that has been out for 3 years at this point, and is not quite as intrusive as #117 that would remove `portalocker` altogether (and in all honesty, that PR has been out there for a year, with no sign of merging).

### DIFF
--- a/msal_extensions/cache_lock.py
+++ b/msal_extensions/cache_lock.py
@@ -6,7 +6,6 @@ import time
 import logging
 
 import portalocker
-from packaging.version import Version
 
 
 logger = logging.getLogger(__name__)
@@ -19,9 +18,6 @@ class CrossPlatLock(object):
     """
     def __init__(self, lockfile_path):
         self._lockpath = lockfile_path
-        # Support for passing through arguments to the open syscall was added in v1.4.0
-        open_kwargs = ({'buffering': 0}
-            if Version(portalocker.__version__) >= Version("1.4.0") else {})
         self._lock = portalocker.Lock(
             lockfile_path,
             mode='wb+',
@@ -30,7 +26,10 @@ class CrossPlatLock(object):
             # More information here:
             # https://docs.python.org/3/library/fcntl.html#fcntl.lockf
             flags=portalocker.LOCK_EX | portalocker.LOCK_NB,
-            **open_kwargs)
+            # Support for passing through arguments to the open syscall
+            # was added in Portalocker v1.4.0 (2019-02-11).
+            buffering=0,
+        )
 
     def _try_to_create_lock_file(self):
         timeout = 5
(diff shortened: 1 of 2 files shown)


akx:
@microsoft-github-policy-service agree

rayluo:
There is a history behind it.

> `packaging` was only used as a dependency to check for the version of `portalocker` in use.
> 
> This drops that dependency and instead corrects the minimum version of `portalocker` to the one checked here.
> 
> It should be safe to depend on a version of `portalocker` that has been out for 3 years at this point

We started with `portalocker=~1.6` in the first place, and then @bluca convinced us to [relax the lower bound for Linux](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/49). If @bluca can chime in to say it is now OK to also use portalocker 1.6+ on Linux, we will consider this PR.



> #117 that would remove `portalocker` altogether (and in all honesty, that PR has been out there for a year, with no sign of merging).

You can chime in in that PR and ping the reviewer there. That PR was developed based on a need at that time, but the integration test effort was suspended, so, there is little motivation to move forward.

akx:
@rayluo Thanks for the feedback. My angle to this is trying to prune extra dependencies that aren't really needed – seems a waste to depend on (and download!) all of `packaging` just to be able to compare a version to another.

If, AIUI, not upgrading Portalocker is mostly the issue of more work for downstream Linux packagers, I understand and we'll need to deal accordingly, packagers can be an ornery bunch 😄 

* It looks like Ubuntu 20.04 (still LTS supported) has `python3-portalocker` `1.5.1-1`, so that would be an issue for pinning to 1.6+.
* Debian Bullseye (the oldest still supported Debian) has `2.2.1`, not an issue.

(I don't quite know how to check for the other major distros.)

Maybe the heuristic for whether the `file_open_kwargs` can be passed could be done without an external dependency – some ideas:
* parse `portalocker.__version__` "by hand" into a `version_info` style tuple and compare it
* check whether `file_open_kwargs` exists in the `__doc__` for `portalocker.Lock.__init__` – a bit hacky, maybe, but should do the trick?

akx:
IOW, @bluca, what do you think, would Portalocker >= 1.5 be a good middle ground 3 years after #49? 😁

bluca:
> IOW, @bluca, what do you think, would Portalocker >= 1.5 be a good middle ground 3 years after #49? 😁

yes, fine by me

akx:
@rayluo Okay, rebased + pushed a version that changes the pin to `>= 1.4`.

rayluo (approved):

PR #126, opened 2023-12-21
Remove mentions of Travis CI

akx:
This repository doesn't seem to have used Travis CI for a while, so there's little point in keeping the configuration files around.


rayluo (approved):
Thanks! We will merge this.

PR #128, opened 2024-04-12
Switch to MSAL 1.27+'s TokenCache._find()

### LINKED ISSUE #127: [Bug] `msal_extensions.token_cache.PersistedTokenCache` is bypassed by `ConfidentialClientApplication`
**Describe the bug**
AzureAD/microsoft-authentication-library-for-python#644 introduced a regression that `msal_extensions.token_cache.PersistedTokenCache` is bypassed by `ConfidentialClientApplication`.

**To Reproduce**
```
az login --service-principal --username ... --password ... --tenant ... --allow-no-subscriptions

az account get-access-token --scope https://management.azure.com//.default
...
  "expiresOn": "2024-04-11 19:25:43.000000",

az account get-access-token --scope https://management.azure.com//.default
...
  "expiresOn": "2024-04-11 19:26:03.000000",
```

Notice each time a new access token is retrieved, bypassing the token cache. Detailed analysis is provided at https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/644#discussion_r1560791416.

**This causes a severe regression in OIDC authentication, so that no Azure CLI task can run longer than the OIDC token's 5-minute lifetime** (https://github.com/Azure/azure-cli/issues/28708#issuecomment-2049400226).

**Expected behavior**
Old access token from the token cache should be retrieved.

**What you see instead**
A new access token is retrieved.

**The MSAL Python version you are using**
1.28.0

**Additional context**

rayluo:
This PR will fix #127 . The fix will be shipped as MSALEX 1.2.0a1 within this month. And then next month both MSAL and MSAL EX will rename the affected function from `_find()` to `iter()`.

CC @jiasli

### DIFF
--- a/docker_run.sh
+++ b/docker_run.sh
@@ -6,9 +6,11 @@ docker build -t $IMAGE_NAME - < Dockerfile
 echo "==== Integration Test for Persistence on Linux (libsecret) ===="
 echo "After seeing the bash prompt, run the following to test encryption on Linux:"
 echo "    pip install -e ."
-echo "    pytest"
+echo "    pytest -s tests/chosen_test_file.py"
+echo "Note that you probably need to set up ENV VAR for the test cases to run"
 docker run --rm -it \
     --privileged \
+    --env-file .env \
     -w /home -v $PWD:/home \
     $IMAGE_NAME \
     $1
--- a/msal_extensions/__init__.py
+++ b/msal_extensions/__init__.py
@@ -1,5 +1,5 @@
 """Provides auxiliary functionality to the `msal` package."""
-__version__ = "1.1.0"
+__version__ = "1.2.0b1"  # Note: During/after release, copy this number to Dockerfile
 
 from .persistence import (
     FilePersistence,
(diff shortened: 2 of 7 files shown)


jiasli:
> And then next month both MSAL and MSAL EX will rename the affected function from `_find()` to `iter()`

You may choose other names as well. This is just an idea that pops up in my mind.

bgavrilMS (approved):

bgavrilMS:
@rayluo  - how to avoid this type of regression in the future? Would it make sense to move the logic in this repo to the main MSAL repo? (we have done this in MSAL Java and MSAL .NET already - it also reduces the number of builds we need to look at).

rayluo:
> @rayluo - how to avoid this type of regression in the future? Would it make sense to move the logic in this repo to the main MSAL repo? (we have done this in MSAL Java and MSAL .NET already - it also reduces the number of builds we need to look at).

I'm afraid merging repo would not be a good option for Python. Nowadays, all interop layers of Python (such as the PyMsalRuntime package) would ideally be implemented with different technology which requires different pipelines anyway. Might as well leave them in a self-contained, separated repo.

Good news is we have lighter weight approach to guard this. Previously I did not anticipate API-level change in token cache (which never happened in the last 5+ years), so I did not setup nightly builds to test it with MSAL's dev branch. My bad. We shall do it this time, and it shall be sufficient.

PR #136, opened 2025-01-11
Allow portalocker version 3

musicinmybrain:
See https://github.com/wolph/portalocker/releases/tag/v3.0.0; portalocker 3 drops support for end-of-life Python versions older than 3.9. The older portalocker versions allowed in the range still cover these Python interpreter versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(
     python_requires=">=3.7",
     install_requires=[
         'msal>=1.29,<2',  # Use TokenCache.search() from MSAL Python 1.29+
-        'portalocker<3,>=1.4',
+        'portalocker<4,>=1.4',
 
         ## We choose to NOT define a hard dependency on this.
         # "pygobject>=3,<4;platform_system=='Linux'",


rayluo:
Thanks for your attempt. Have you also tested it with portalocker 3.0?

On the other hand, we intend to go another direction and remove portalocker altogether. You may want to help us test #117 instead.

musicinmybrain:
> Thanks for your attempt. Have you also tested it with portalocker 3.0?

Well, I used `tox` to run the tests, and they all still passed. I don’t have any other ways to test this.

> On the other hand, we intend to go another direction and remove portalocker altogether. You may want to help us test #117 instead.

It would seem that this PR still makes sense (though it will need to be rebased) even after https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/117 moves the portalocker dependency to an extra.

Again, I unfortunately don’t have anything to authenticate against or any particular experience with this library, and my ability to test this is limited to running `tox`.

This PR is motivated by the need to accommodate portalocker 3 in [Fedora’s `python-msal-extensions` package](https://src.fedoraproject.org/rpms/python-msal-extensions).

musicinmybrain:
Note that dropping support for Python 3.9 is the *only* documented reason for portalocker bumping the major version, and the source diff https://github.com/wolph/portalocker/compare/v2.10.1...v3.0.0 seems to agree; the changes look really minor.

PR #146, opened 2025-08-06
Update minimum version of portalocker

kurtqq:
This helps avoid ResolutionTooDeepError  with pip when it starts backtrack to search for matched versions. This happens in large projects that uses many libs. The easiest way to avoid it is to increase minimum versions


rayluo:
1. Out of curiosity, can you provide a url or something pointing us to a log with that `ResolutionTooDeepError`?
2. More importantly, starting from [msal-extensions version 1.3](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/releases/tag/1.3.0), this library does not use portalocker by default. We prefer phasing out that dependency, rather than sending a misleading signal of keeping it up-to-date. Existing customers shall not even notice its removal from the default installation. (That is also why we are curious to know why you ran into `#1` above in the first place.)

